PR #151, opened 2017-09-01
Ignore derived ID attributes

### LINKED ISSUE #150: Support Derived IDs
[AtomicEChangeResolver@131](https://github.com/vitruv-tools/Vitruv/blob/master/bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/resolve/AtomicEChangeResolver.xtend#L131) throws an exception if there’s an ID attribute that’s not changeable. But that should not happen if the attribute is derived.

As discussed in #146, the solution should be to not store values of derived ID attributes in changes.

HeikoKlare:
This is an optimized solution for the problem mentioned in #150 and initially solved in #146.

With this PR, changes do not store IDs for ID attributes that are derived, as it makes no sense to store derived information and to (try to) apply them on model elements when applying the change.

This PR fixes #150.


jGleitz:
Thanks for making these changes!

Issue #131, opened 2017-08-22
Support addition, retrieval and removal of m2n-correspondences in reactions


max-kramer:
Support m2n-correspondences in reactions. They are needed e.g. to realize m2n-mappings as in the addresses-recipients demo and are already internally supported but currently partially emulated using tags.

jGleitz:
As I see it, `m:n` is already fully supported for fixed `m` and `n` using tags.

What’s not supported is `*:n` or `*:*`. I have a use case for these cases, too. From my point of view, those could be tackled by introducing a `many` keyword for correspondence retrieval: 
```
val x = retrieve many y corresponding to z (tagged with a)?
```

which would return a list with all such correspondences. With this, `*:n` would be supported. 

Additionally, one could add type restrictions on correspondences (instead of instance restrictrions), e.g. like this

```
val x = retrieve (many)? y corresponding to instanceof z (tagged with a)?
```

wher `z` would be a Metaclass reference. With this, `*:*` could be realised.

As @max-kramer said, both cases are already supported by the API.

HeikoKlare:
The many retrieval is already discussed in #36 and will be implemented like proposed by you soon.
Nevertheless, many retrievals only solve half the problem, because they also have to be created. Many retrievals are not necessarily used for retrieving 1:n correspondences but also for retrieving multiple 1:1 correspondences.

Regarding *:* correspondences: We first need a use case that show the requirement for such a feature. From my (naive) point of view, such a correspondence should never be necessary, but a concrete use case may convince me of the contrary.

jGleitz:
> Nevertheless, many retrievals only solve half the problem, because they also have to be created Many retrievals are not necessarily used for retrieving 1:n correspondences but also for retrieving multiple 1:1 correspondences.

I’m not quite sure what the difference between a `1:n` correspondence and many `1:1` correspondences with an equal element on one side should be. I actually only thought about the latter case. The advantage is that you can add many `1:1` correspondences without any additional syntax.

Regarding `*:*` correspondences, I have a use case were I’d like to get all correspondences with one type on one side and another type on the other side. But as I’m using the reactions to generate code and not in a “natural” way, that may be a bad example. I get your point that you first want to have a vialid use case.

HeikoKlare:
I would be very interested in that case in which `*:*` correspondences are needed, as you mention.

[...]

PR #155, opened 2017-09-01
Optimized Reactions retrieval

### LINKED ISSUE #36: Allow collection retrieval in Reactions
Reactions do only allow the retrieval of a single element in the matcher.
Sometimes it is necessary to retrieve a collection of elements, as it is only know that an element corresponds to an arbitrary number of elements of a certain type. For example, in the families-persons scenario, a person in the persons modelcan be member of several families in the families model (e.g. one son and several fathers). Assuming that it is allowed to be father in several families, it is not possible to restrict the retrieval to a single member with the "with"-condition. 

We have to possibilies to integrate such a retrieval into the reactions:
1. We allow to specify "retrieve collection of ..." statements that write the collection of elements to the specified variable, which can be processed within the routine.
2. We allow to specify a single "retrieve collection of..." statement per routine, which results in the execution of the routine for each of the elements in the collection.

Which behavior is more reasonable? The second one assumes that a consistency relation does only contain a static number of elements, thus of you retrieve a collection, each of the elements in the collection belongs to a single consistency relation, which you can restore in one routine execution. The first possibility is more flexible and should thus be only used if we think that it is possible to have one consistency relation which all these elements belong to, so that it is necessary to be able to reference all of them in one routine execution.

### LINKED ISSUE #143: Asserted retrievals in Reactions
Reactions currently support an ordinary element retrieval from the correspondence model, as well as an optional one. While the former retrieves and element and aborts the routine execution if there is no element, the latter returns null if there is no corresponding element.

Since in some cases, the retrieval should not determine if the routine has to be executed or not, but is mandatory because the consistency specifications require the correspondence to always exist when reaching the routine, it would be helpful to have a definition of this expected behavior so that testing and debugging is more easy.

We therefore propose a kind of "asserted" retrieve, which leads to an exception if the retrieve operations fails.
Finally, this will lead to three kinds of retrieval, for which we have to find concise names, such as
- `retrieve asserting`
- `retrieve matching`
- `retrieve optional`

**We would be glad to get proposals for better keywords.**

The first to properties will also be attached to check operations in a Reactions matcher.

HeikoKlare:
This PR improves the retrieval operations in Reactions. It is now possible to define _many_ retrievals for retrieving collections of elements. Furthermore, extending `retrieve` with `asserted` defines that the correspondences is asserted to exist, resulting in an exception if it cannot be resolved. The latter extension is also added for `check` statements.

Fixes #36, fixes #143 and contributes to #131

### DIFF
--- a/bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
+++ b/bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
@@ -313,11 +313,13 @@ class FluentRoutineBuilder extends FluentReactionsSegmentChildBuilder {
 		}
 
 		def check(Function<RoutineTypeProvider, XExpression> expressionBuilder) {
-			routine.matcher.matcherStatements ReactionsLanguageFactory.eINSTANCE.createMatcherCheckStatement => [
+			val statement = ReactionsLanguageFactory.eINSTANCE.createMatcherCheckStatement => [
 				code = XbaseFactory.eINSTANCE.createXBlockExpression.whenJvmTypes [
 					expressions += extractExpressions(expressionBuilder.apply(typeProvider))
 				]
 			]
+			routine.matcher.matcherStatements += statement
+			return statement
 		}
 		
 		def checkAsserted(Function<RoutineTypeProvider, XExpression> expressionBuilder) {
--- a/bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
+++ b/bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
@@ -26,6 +26,8 @@ import static tools.vitruv.dsls.reactions.codegen.helper.ReactionsLanguageConsta
 import tools.vitruv.dsls.reactions.reactionsLanguage.RetrieveOneModelElement
 import tools.vitruv.dsls.reactions.reactionsLanguage.RetrieveOneElementType
 import tools.vitruv.dsls.reactions.reactionsLanguage.RetrieveOrRequireAbscenceOfModelElement
+import tools.vitruv.dsls.reactions.reactionsLanguage.MatcherCheckStatement
+import tools.vitruv.dsls.reactions.reactionsLanguage.CheckType
 
 class FluentRoutineBuilder extends FluentReactionsSegmentChildBuilder {
 
@@ -324,12 +326,32 @@ class FluentRoutineBuilder extends FluentReactionsSegmentChildBuilder {
 			return new RetrieveModelElementMatcherStatementCorrespondenceBuilder(builder, statement)
 		}
 
-		def void check(Function<RoutineTypeProvider, XExpression> expressionBuilder) {
-			routine.matcher.matcherStatements += ReactionsLanguageFactory.eINSTANCE.createMatcherCheckStatement => [
+		def check(Function<RoutineTypeProvider, XExpression> expressionBuilder) {
+			val statement = ReactionsLanguageFactory.eINSTANCE.createMatcherCheckStatement => [
 				code = XbaseFactory.eINSTANCE.createXBlockExpression.whenJvmTypes [
 					expressions += extractExpressions(expressionBuilder.apply(typeProvider))
 				]
 			]
+			routine.matcher.matcherStatements += statement
+			return new MatcherCheckTypeBuilder(builder, statement);
+		}
+	}
+	
+	static class MatcherCheckTypeBuilder {
+		val extension FluentRoutineBuilder builder
+		val MatcherCheckStatement statement
+
+		private new(FluentRoutineBuilder builder, MatcherCheckStatement statement) {
+			this.builder = builder
+			this.statement = statement
+		}
+
+		def void asserted() {
+			statement.checkType = CheckType.ASSERTED
+		}
+
+		def void matching() {
+			statement.checkType = CheckType.MATCHING
 		}
 	}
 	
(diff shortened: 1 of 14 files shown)


jGleitz:
Also fixes #131, doesn't it?

HeikoKlare:
It fixes only parts of #131. What is still missing is the creation of m:n correspondences and even the retrieval is incomplete. One goal of #131 is to retrieve correspondences for a collection of elements, which is currently not possible.
Nevertheless, this PR contributes to that issue.

bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
```diff
@@ -282,20 +307,23 @@ class FluentRoutineBuilder extends FluentReactionsSegmentChildBuilder {
...
-			]).reference(absentMetaclass)
+			val statement = ReactionsLanguageFactory.eINSTANCE.createRequireAbscenceOfModelElement.reference(absentMetaclass)
 			routine.matcher.matcherStatements += statement
 			return new RetrieveModelElementMatcherStatementCorrespondenceBuilder(builder, statement)
 		}
 
-		def void check(Function<RoutineTypeProvider, XExpression> expressionBuilder) {
-			routine.matcher.matcherStatements += ReactionsLanguageFactory.eINSTANCE.createMatcherCheckStatement => [
+		def check(Function<RoutineTypeProvider, XExpression> expressionBuilder) {
+			routine.matcher.matcherStatements ReactionsLanguageFactory.eINSTANCE.createMatcherCheckStatement => [
```
HeikoKlare: Here is something missing

bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
```diff
@@ -212,24 +214,47 @@ class FluentRoutineBuilder extends FluentReactionsSegmentChildBuilder {
...
+		}
+
+		def void retrieveMany(EClass modelElement) {
+			reference(modelElement)
+			statement.retrievalType = ReactionsLanguageFactory.eINSTANCE.createRetrieveManyModelElements();
+		}
+
+		def void reference(EClass modelElement) {
+			statement.reference(modelElement)
 		}
```
jGleitz: Is there any reason to make this method public?

bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
jGleitz: * Why did you make the method return something?
 * Why did you remove the `+=`?

bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
```diff
@@ -212,24 +214,47 @@ class FluentRoutineBuilder extends FluentReactionsSegmentChildBuilder {
...
+		def retrieveAsserted(EClass modelElement) {
+			val retrieveOneStatement = internalRetrieve(modelElement);
+			retrieveOneStatement.asserted = true
+			return new RetrieveModelElementMatcherStatementCorrespondenceBuilder(builder, statement)
+		}
+
+		def void retrieveMany(EClass modelElement) {
+			reference(modelElement)
+			statement.retrievalType = ReactionsLanguageFactory.eINSTANCE.createRetrieveManyModelElements();
+		}
```
jGleitz: Shouldn’t we continue with a Builder similar to `RetrieveModelElementMatcherStatementCorrespondenceBuilder` (but taking an `EClass` instead of an `EObject`) here?

HeikoKlare: Yes, this is not correct. This mistake was introduced during my last refactorig. I will fix it.

bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
HeikoKlare: No, I just forgot to add the modifier.

bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
HeikoKlare: I fixed this in the last commit on this PR. This review is based on the second last revision.

bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
HeikoKlare: Btw: taking an `EObject` is still correct, as many retrievals can also be performed on model elements. Restricting this to `EClass` would only allow to retrievel correspondences on metamodel elements, which is also supported, but which the implementation is not restricted to.

jGleitz: I see. So adding

```Xtend
return new RetrieveModelElementMatcherStatementCorrespondenceBuilder(builder, statement)
```

to the method should be sufficient.

bundles/dsls/tools.vitruv.dsls.reactions/src/tools/vitruv/dsls/reactions/builder/FluentRoutineBuilder.xtend
jGleitz: Oh, my bad.

PR #239, opened 2018-10-12
Correspondence model refactoring

HeikoKlare:
This PR completely refactors the correspondence model. The internal implementation is now completely encapsulated in the `InternalCorrespondenceModelImpl` and views, typed with the correspondence type, are strictly separated from the internal representation. The internal representation does not work as a generic view any more. This is necessary to allow the views to provide different functionality than the internal representation. E.g. the internal representation has to expect a correspondence type for `getCorrespondingEObjects`, as it internally has to get the appropriate correspondendes to extract the elements from. To hide this from the user, the view interface provides functions without the necessity to specify the correspondence type, which is automatically set by the type of the view.
This allows to remove redundancies by implementing functionality for getting corresponding elements based on getting the existing correspondences.
Additionally, a factory for creating the correspondence model is introduced to avoid the necessity to expose the implementation class.

This PR also fixes a bug in the `getCorrespondences` operation of the `CorrespondenceModelImpl`, which returned only TUID-based correspondences. Logic for also extracting UUID-based correspondences was added.


HeikoKlare:
Closed due to newer PR #243

PR #260, opened 2019-01-17
Unused Import Removal (Xtend Update)

HeikoKlare:
Remove import that is unsed because of an update of Xtend already providing the extension method.

### DIFF
--- a/bundles/extensions/dslsruntime/tools.vitruv.extensions.dslsruntime.commonalities/src/tools/vitruv/extensions/dslruntime/commonalities/resources/impl/IntermediateResourceBridgeI.xtend
+++ b/bundles/extensions/dslsruntime/tools.vitruv.extensions.dslsruntime.commonalities/src/tools/vitruv/extensions/dslruntime/commonalities/resources/impl/IntermediateResourceBridgeI.xtend
@@ -15,8 +15,6 @@ import tools.vitruv.framework.util.datatypes.VURI
 
 import static com.google.common.base.Preconditions.*
 
-import static extension edu.kit.ipd.sdq.commons.util.java.lang.IterableUtil.*
-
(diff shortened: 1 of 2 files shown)


cleanTargetFolder.sh
```diff
@@ -0,0 +1 @@
+find -type d -name target -prune -exec rm -rf {} \;
```
jGleitz: * There is a maven command for this, isn’t there: `mvn clean`?
 * Is this really necessary? In the past, I cloud always solve this error by setting up the Eclipse project correctly.

HeikoKlare: * `mvn clean` does also remove the `src-gen` folders with the generated code.
* I really hope this is not necessary :-) I tried to find out how to correctly set up the Eclipse projects, but I did not find a way rather than explicitly ignoring `target/` resources in the project settings or manually setting all `target` folders to _derived_. Otherwise the `target` folders are always somehow treated as source folders, although they are not declared as such and although Maven is not used within Eclipse. Using Maven within Eclipse is also no solution because not all projects can be treated as Maven projects (as they have no `pom.xml`) and because some of our Maven plugins are not supported by m2e. What is your way of setting up the Eclipse projects to avoid problems?

HeikoKlare: You are completely right. I must have made a mistake when trying to exclude the `target` folder in Eclipse since it properly works now (and whereas script was just a hack, I think excluding the folder is the right way to go). If someone imports the projects as Maven projects into Eclipse this may be problematic, but since m2e does not support all of our plugins this does not seem like an issue to me right now.

HeikoKlare:
Remark: The forced push simply removed a script for cleaning `target` folder that I added before, but which is obsolete due to a different (better) solution. I removed the commit to keep the history clean.

PR #263, opened 2019-01-23
Remove SimpleTextXBlockExpression

HeikoKlare:
This PR removes the class `SimpleTextXBlockExpression`, which was a wrapper for simple text to be passed as code blocks in the Reations language. Since the language itselfs expectes an `XBlockExpression`, no text fragments could be passed, if the Reaction was not created from a textual artifact but was instantiated programmatically.
The `SimpleTextXBlockExpression` therefore extended the `XBlockExpression` and redefined the `getText()` operation to simply return the text it was initialized with. This broke several API restrictions and led to the necessity to always make a type check in the Reactions language compiler whenever an `XBlockExpression` was used. This is really bad practice and since there is currently no usage of the class, because all programmatic instantiations of Reactions are now done with the `FluentReactionBuilder` (thanks to @jGleitz), we remove this class now to clean up the code.


jGleitz:
:tada: :clap:

PR #296, opened 2020-03-22
Eclipse 2020-03

HeikoKlare:
This PR makes some adjustments for Eclipse 2020-03, which should be merged when moving all Vitruv projects to Eclipse 2020-03 with JUnit 2.13 and Xtext 2.21.
Namely, it removes some imports in Xtend files that are obsolete and marked as warnings in current Xtend versions. Additionally, it corrects some JUnit tests cases expecting exceptions, for which JUnit 2.13 provides a new API (and one we use is deprecated).


HeikoKlare:
This PR is obsolete due to #302 upgrading to Eclipse 2020-06 and #303 upgrade Junit to 2.13.

PR #327, opened 2020-12-04
Release 1.0.0

HeikoKlare:
This PR performs a delayed framework release for a state before significant changes especially to the test infrastructure (in September 2020).
I propose to finally have a first major release. Since this is a research prototype, we will probably not come to a state that we could call "post-alpha" quite soon.
We can also think about adapting the major version properly after breaking API changes in the future, even if this leads to having major version increments quite often.


jGleitz:
> We can also think about adapting the major version properly after breaking API changes in the future.

:+1: :+1: :+1:! Releasing more often and taking semantic versioning seriously would avoid breaking applications when making changes to the Framework.

PR #330, opened 2020-12-09
Avoid duplicate resource creation in tests

HeikoKlare:
The current implementation of resource loading in tests leads to duplicate resources, because `resourceSet.getResource` creates a `Resource` even if it throws an exception. Calling `createResource` afterwards leads to two resources with the same URI in the `ResourceSet`. This is probably hidden by clearing the `ResourceSet` after each change propagation, such that resources are properly reloaded anyway, but still this is a bug and leads to problems when implementing a fallback propagation mechanism that does not clean the `ResourceSet` after each change propagation.


bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/VitruvApplicationTest.xtend
```diff
@@ -93,13 +92,10 @@ abstract class VitruvApplicationTest implements CorrespondenceModelContainer {
 	 * @param modelPathWithinProject A project-relative path to a model.
 	 */
 	def protected Resource resourceAt(Path modelPathWithinProject) {
-		try {
+		if(!resourceSet.getURIConverter().exists(getPlatformModelUri(modelPathWithinProject), newHashMap)) {
```
HeikoKlare, line 95: If forgot to move the existence check into  a separate method. I will improve that.

jGleitz:
> because resourceSet.getResource creates a Resource even if it throws an exception.

are you kidding me :open_mouth:? Sometimes EMF feels like it breaks sensible expectations on purpose :roll_eyes:.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/VitruvApplicationTest.xtend
jGleitz: `newHashMap` should be `emptyMap`, or, even simpler, `null`.

jGleitz (changes requested):

jGleitz:
Technically, this code is still not correct, because it does not handle a resource vanishing between the existence check and loading it (you should never check whether some shared resource exists before accessing it, but rather handle errors during access).

I guess the correct way to load an EMF resource would be (very intuitively!):

```Xtend
try {
	loadModelResource(modelPathWithinProject)
} catch (RuntimeException e) {
	if (e.cause instanceof FileNotFoundException) {
		unloadModelResource(modelPathWithinProject)
		createModelResource(modelPathWithinProject)
	} else throw e
}
```

right?

HeikoKlare:
Actually, the solution that you propose is comparable to my first attempt of solving the problem.
There are, however, two important things to distinguish, which are resource creation and resource loading.
Creating a resource is about adding a `Resource` instance with a specific URI to a `ResourceSet`.
Loading a resource is about deserializing a resource persistence from the file system.
Finally, unloading is only about removing the loaded resource contents, not about removing a `Resource` from a `ResourceSet`.

When we talk about shared resources, we need to be sure which kind of shared resource we mean. You seem to talk about the files as the persistence of resources to be the shared resource (with resource having a different meaning than the EMF `Resource`).
You would, however, need to lock the file system to avoid current modifications of that kind of resource.
Additionally, these files only becomes changed by either the test (in the test `ResourceSet`), or by the V-SUM during change propagation.
Such changes do, however, never occur at the same time. So there should be no risk of concurrent access to such a file. 
In the future, views should even not be persisted (at least not at the same place where the V-SUM file is placed) and then there is even no problem of concurrency between views and V-SUM.
The other part of resource sharing is about concurrent access to the test `ResourceSet`. We can easily solve that by synchronizing the `resourceAt` method (or `loadModelResource` or whatever accesses it in a transaction) on the `ResourceSet`.

All in all, I do not understand the problem (or at least do not understand how to solve it), because your proposed solution does not seem to solve the problem either, as between loading, unloading and creation, changes in the file system can still be performed.
In my opinion, we cannot exclude that concurrent changes at the file system can be made, except for the (implicit) protocol defined by the interaction between V-SUM and views that currently ensures it.

One further implementation option is to simply let EMF try to load the `Resource` (with `resourceSet.getResource(..., true)`), catch the exception and unload any inconsistent state in case of a failure, because this creates the resource and only fails during loading if not persistence at the resource URI exists.
In my opinion this is, however, worse than the proposed solution, because you need to understand what EMF does during resource loading (and that is basically the same that we do in my proposed correction explicitly) instead of having it presented explicitly.
It could look something like this (and you can even remove the first get and the create, but then have to extract the resource after the failing operation):
```
var resource = resourceSet.getResource(getPlatformModelUri(modelPathWithinProject), false)
if (resource === null) {
	resource = resourceSet.createResource(getPlatformModelUri(modelPathWithinProject))
	try {
		resourceSet.getResource(getPlatformModelUri(modelPathWithinProject), true)
	} catch (RuntimeException e) {
		resource.unload
	}
}
```

To be honest, I am not fully convinced of any of the discussed solutions yet. So, of course, I am still open for proposals.

HeikoKlare:
> > because resourceSet.getResource creates a Resource even if it throws an exception.
> 
> are you kidding me 😮? Sometimes EMF feels like it breaks sensible expectations on purpose 🙄.

Well, I do not agree to that. If EMF would not create a resource if you set `loadOnDemand` to `true`, than you could not have expected the code you added before to work ;-) But maybe this becomes clearer with my previous comment on what resources actually are.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/VitruvApplicationTest.xtend
HeikoKlare: Absolutely. I will change that as soon as we agree to some general solution :-)

jGleitz:
Wow, so much to untangle here. Let’s go.

> we need to be sure which kind of shared resource we mean. […] You would, however, need to lock the file system to avoid current modifications of that kind of resource.

Let’s make sure to distinguish the term ‘resource’ (the English word and technical term) and `Resource` (this weird EMF thing). What I was talking about was a basic programming principle: Never check for the existence of a shared resource, but rather access the resource and then handle the case that it does not exist. Checking beforehand is an antipattern because you’ll need to handle the case that the resource vanishes anyway. So not checking but rather handling the error should always be prefered.

>  Such changes do, however, never occur at the same time. So there should be no risk of concurrent access to such a file.

That’s true for the current state of affairs. That’s why I said, ‘_Technically_, this code is still not correct’. It’s not an important issue, but still a good pattern to follow.

> One further implementation option is to simply let EMF try to load the Resource (with resourceSet.getResource(..., true)), catch the exception and unload any inconsistent state in case of a failure, because this creates the resource and only fails during loading if not persistence at the resource URI exists.

Exactly! This is exactly what I wanted to propose with my [comment](https://github.com/vitruv-tools/Vitruv/pull/330#issuecomment-741739974). However, I see now that the `createModelResource(modelPathWithinProject)` in the `catch` clause is superfluous.

> because you need to understand what EMF does during resource loading 

Well yeah, if you have an unintuitive API, you are going to have unintuitive code.

>  If EMF would not create a resource if you set loadOnDemand to true, then you could not have expected the code you added before to work 

Why? My code called `createResource` if loading the resource failed! I think it is fair to assume that an operation has no side effect if it fails for a predictable reason.

If I request a resource with the additional condition to load it from the file system if it does not yet exist, and the resource does not exist but can also not be loaded from the file system, I really don’t think creating an empty virtual `Resource` is a valid response. The thing I asked for is not possible, creating a virtual `Resource` is nothing I would expect. But hey, what does it matter, EMF is the way it is.


To conclude, I think this is the most sensible code:

```Xtend
val resourceUri = getPlatformModelUri(modelPathWithinProject)
return try {
	resourceSet.getResource(resourceUri, true)
} catch (RuntimeException e) {
	if (e.cause instanceof FileNotFoundException) {
 		// EMF has created an empty resource after failing to load it from the file system
		resourceSet.getResource(resourceUri, false)
	}
}
```

From my understanding, we don’t even need to unload the resource. If that’s wrong, then we should also call  `unload` in the `catch` clause, of course.

```

**Comment by HeikoKlare** [comment]
In short: I understand and agree to your points. Let's make the best out of the strange EMF behavior.

I have proposed a solution, which only slightly derives from our one:
1. I have `synchronized` the method on the `resourceSet` to ensure that the whole method runs as a transactions on the `ResourceSet` (and because acces to a `ResourceSet` does not seem to be synchronized at all).
2. I have replaced the check for the exception cause with a check of resource existence and added the creation of the resource in case if it does not exist. This is for two reasons: First, `FileNotFoundException` is only thrown for file URI. Using platform URIs a `ResourceException` is thrown, which is not accessible at all. Second, this seem to be the most safe implementation to me, because independent from the causing error and independent from the strange behavior of `getResource` (which does not run as a transaction, as you have pointed out), we should be able to deliver a proper `Resource` with that solution. I did not add an explicit `unload` because at least the loading seems to run as a transaction within EMF.

**Comment by HeikoKlare** [comment]
Since we seem to conclude and since I need these changes for properly proceeding with adaptations of the applications, I will merge them. If there are still points of criticism or discussion, feel free to add them and we can still solve them with further pull requests.

File: bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/VitruvApplicationTest.xtend [hunk 2]
```diff
@@ -93,13 +92,19 @@ abstract class VitruvApplicationTest implements CorrespondenceModelContainer {
...
+				resource = getAndLoadModelResource(modelPathWithinProject)
+			} catch (RuntimeException e) {
+				// EMF failed during demand creation, usually because loading from the file system failed.
+				// If it has created an empty resource, retrieve it, and otherwise create one.
+				resource = getModelResource(modelPathWithinProject)
+				if (resource === null) {
+					resource = createModelResource(modelPathWithinProject)
+				}
+			}
+			return resource
```
jGleitz, line 107: 1. I don’t think removing the `if (e.cause instanceof FileNotFoundException)` is sensible. Other kinds of errors (say file permission errors) should be thrown. Without the check, the implementation silently (!) creates an empty `Resource` even if the target resource exists on the file system, but cannot be accessed. This is certainly not expectable behaviour, from my point of view.
2. In which case will `getModelResource` return `null` after an exception? If we don't know whether it can happen, I’d suggest doing something like `checkNotNull(getModelResource(modelPathWithinProject), "EMF did not create the Resource after not finding the source file!")` to learn whether and why it can happen.
2. Synchronizing on `resourceSet` seems unnecessarily expensive to me. Does `ResourceSet` not support loading different resources in parallel. If I understand the situation correctly, we can remove the synchronization if we remove the `createModelResource` call (see previous point). Otherwise, I’d suggest resolving the URI and synchronize on that.
2. If you want, the code could be simplified to:
```Xtend
synchronized (resourceSet) {
	return try {
		getAndLoadModelResource(modelPathWithinProject)
	} catch (RuntimeException e) {
		if (e.cause instanceof FileNotFoundException) {
			getModelResource(modelPathWithinProject) ?: createModelResource(modelPathWithinProject)
		} else throw e
	}
}
```

jGleitz (changes requested):

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/VitruvApplicationTest.xtend
HeikoKlare: 1. As I mentioned before, catching a `FileNotFoundException` is not sufficient.
2. The documentation of `getResource` says that resource creation can also throw a `RuntimeException`, thus there will be cases in which this can happen.
3. Taking a look into the EMF implementation, I would expect EMF to fail when running two `getResource` calls in parallel, especially when one resolved proxies that are contained in the other resource. In my opinion it is better to have this synchronization (since we do not perform parallel resource loadings and thus it does not produce overview) instead of failing due to concurrency issues if we, by accident, try to load resources in parallel.

jGleitz: 1. Sorry, I missed your other comment. Then we should find another solution, swallowing all other errors and pretending the file does not exist, even if it might exist, is not acceptable from my point of view. This hides bugs.
2. If resource creation throws a `RuntimeException`, then `getResource` will return `null`? I’d expect the exception to be rethrown. Also: Maybe we don’t want to continue after such a `RuntimeException`?

HeikoKlare: 1. I found the risk to miss some exception time (as we missed `ResourceException` before) to be higher and more severe than swallowing potential other exceptions, but you are right that we should find a better and finally robust solution.
2. I do not get it, but maybe it's just too late for me. The first time we try to load a resource (`getResource(..., true)`) may throw a `RuntimeException`, because the resource cannot be demand created. Then, in the `catch` block, calling `getResource(..., false)` will return `null`, as there is no resource with the given `URI` and none will be demand created and loaded (so no exception should be thrown). In that case, we create the `Resource`.
3. Maybe, yes. I was actually thinking about the Java2PCM tests, in which monitors asynchronously trigger change propagation, but they do not touch the `ResourceSet` in the test classes.

jGleitz: 1. Yes, I see that you assessed the risk that way. However, I think this is a bad assessment: If we miss some exception, it will be thrown in our face, we can look at it and potentially add it to the `if`-clause. Easy. If, on the other hand, we swallow an exception (for example because of access rights), an empty resource will wrongfully be created and it might take hours to figure out what is going on (especially for somebody not familiar with the code, like us in 3 months). That’s bad!

I am insisting on this point that much because it is a pattern I have seen time and time again in the Vitruv code base: Exceptions get swallowed a lot, and the general sentiment seems to be ‘better catch too much than having another execption’. I think that that is a _fundamentally_ wrong approach.

> The first time we try to load a resource (`getResource(..., true)`) may throw a `RuntimeException`, because the resource cannot be demand created. Then, in the catch block, calling `getResource(..., false)` will return `null`, as there is no resource with the given `URI` and none will be demand created

I thought the whole point of this PR was that that is _not_ what is happening? You wrote it yourself, in the PR description: 

> `resourceSet.getResource` creates a `Resource` even if it throws an exception

HeikoKlare: Well, I totally agree. I think I was confused, because some proposal that we have discussed did not rethrow the exception (and would thus have also swalled other exceptions) and I think I missed that you proposed to do so in the last proposal. That is, of course, a better solution. I am just not sure yet how to properly identify the `ResourceException`, as it seems to be part of an internal API, but I will figure that out.

Regarding `getResource`: The behavior depends on the argument `loadOnDemand`, as mentioned by the documentation. Only if this is set to `true`, a `Resource` will be created and demand loaded (which fails if no persistence exists at that `URI`). If you set this to `false`, you only get the `Resource` already present in the `ResourceSet`, if existing, and otherwise `null`. My initial statement seems to be insufficient in this regard, because I did not point out that this behavior does not occur with `loadOnDemand=false`. I am sorry for that.

HeikoKlare: Just as a note for myself: We should apply the solution at some other places as well. Especially `ResourceRepositoryImpl` has several of these implementations, which should be fixed as well.
In `ProjectBuilderUtils`, exceptions due to these implementations are swallowed by not rethrowing them but only submitting a fatal log entry.

jGleitz: Regarding `getResource`: Yes, this is my whole point: We call `getResource` with `loadOnDemand=true` in the `try`-clause. This creates a `Resource` even if it throws an exception. Hence, when we call `getResource` with `loadOnDemand=false` in the `catch`-clause we are guaranteed to get a resource, aren’t we?

HeikoKlare: > This creates a Resource even if it throws an exception.
Unfortunately not. `getResource` with `loadOnDemand=true` can even fail during resource creation and then throws a `RuntimeException`. In that case, we end up in the `catch` clause and `getResource` with `loadOnDemand=false` will return `null`.

jGleitz: Yes, but will this happen if the exception’s cause is that the resource does not exist? All other exceptions should be rethrown anyway.

HeikoKlare: No, I would not expect that. Sorry, my answer concerned the current solution (without checking the cause of the exception) and not a solution with a cause check. In that case, simply creating the resource without trying to retrieve it should be sufficient.

There still seems to a problem: The `ResourceException`, which occurs when no file persistence of a resource with platform URI is found, is, first, internal, and, second, can also occur when some other failure during loading happened. So it is not sufficient to identify the absence of a file.
I do not really know how to properly deal with the whole situation, because investigating the exception does not seem to be a suitable solution. What if we try to load the resource, just like we do now, and if we catch an exception, we validate whether the file exists and, if not, create a resource? Otherwise, we rethrow the exception. The worst that could happen is that the file is somehow generated between the failed load and the existence check, right? This means that we may fail more often than we would do when properly investigating the exception, but to me that seems to be irrelevant.

jGleitz: > In that case, simply creating the resource without trying to retrieve it should be sufficient.

Okay. Glad that we agree.

 ---

The fact that EMF gives us no chance to see what went wrong by inspecting the exception is unbelievable. Does the `ResourceException` have a message we can recognize? I know that this solution would rely on an implementation detail, but it would still be an easy solution.

If that is not an option, I’d propose to implement the heuristic you propose, but a little more ‘clever’:

 1. Check whether the exception is a `FileNotFoundException`
 1. If not, check whether the exception’s class’ fully qualified name is that of the (not accessible) `ResourceException`.
 1. If this is the case, check whether the file exists.
 1. In all other cases, rethrow the exception.

This is all ugly, of course, but I guess it’s the best heuristic we can do.

HeikoKlare: The message of the `ResourceException` contains `not exist`. An implementation that works thus may look as follows:
```Xtend
try {
	return resourceSet.getResource(modelUri, true)
} catch (RuntimeException e) {
	// EMF failed during demand creation, usually because loading from the file system failed.
	// If it has created an empty resource, retrieve it, and otherwise create one.
	if (e.cause instanceof FileNotFoundException || e.message.contains("not exist")) {
		return resourceSet.createResource(modelUri)
	} else {
		throw e;
	}
}
```

Do you agree with that solution? In that case I would implement it as a central utility method and use it as a replacement for all the different implementations of resource loading that we have in the framework.

jGleitz: Yes, I agree with this one :+1:

Only thing: Please remove ‘and otherwise create one’ from the comment, since it is not accurate any more.

HeikoKlare: Yes, thank you! And more importantly, it should be `return resourceSet.getResource(modelUri, false)` rather than `return resourceSet.createResource(modelUri)` in the implementation, because otherwise we reimplement the original mistake ;-)

jGleitz: True :laughing:

HeikoKlare: This should be finally resolved by #339.

PR #331, opened 2020-12-09
Correct identification of resource of a change

HeikoKlare:
By now, we considered the `Resource` of any of the `EObjects` within a change to represent the `Resource` of that change.
This can lead to returning the `Resource` of a referenced `EObject`, such as a primitive type in a dedicated repository.
This was not intended and does not match the determination of the `URI` of the change, which does only consider the affected and not referenced `EObjects` of a change.
Since the logic is already defined within the `getURI` method, we can reuse it to identify the proper resource.


jGleitz:
I see. Some feedback: I find the name ‘affectedEObject’ misleading if it has the semantics you say it has. For me, the ‘affected’ objects of a change are the objects that have been changed. With its current semantics, I think the name should rather be something like ‘involvedEObjects’.

jGleitz:
The more I think of it, the less the semantics make sense: A change may have multiple `affectedEObjects`, that may come from different Resources, but a change has only one `affectedResource`. This is counter-intuitive, to say the least.

My proposal: rename `affectedEObjects` to `involvedEObjects`, introduce `changedEObjects` and rename `affectedResource` to `changedResource` with the invariant: `∀ o ∈ changedObjects: o.eResource == changedResource`.

HeikoKlare:
You are right. I also found the name of the method `getAffectedEObjects` very confusing. I already took a note and will post an issue for that to be resolved after adapting the application. I first want to achieve a framework state that works properly with the applications instead of messing up the adaptation with further renamings.

Regarding `changedResource`: I like that proposal and will perform the renaming. We just added the `affectedResource` method and only use it within the `VitruvApplicationTest`, so changing it should be easy and not affect any further code.

HeikoKlare:
However, `affectedResource` conforms to what `affectsEObjects` actually are: those of which attributes or references are changes. The only mistake here is the method `getAffectedEObjects` because it does not only return the `affectedEObjects` of changes. Thus, we could also keep `affectedResource`.

jGleitz:
> However, affectedResource conforms to what affectsEObjects actually are: those of which attributes or references are changes. The only mistake here is the method getAffectedEObjects because it does not only return the affectedEObjects of changes. Thus, we could also keep affectedResource.

That is true. However, my intention was to completely do away with the `affected` lingo. I propose this for two reasons:

 1. `affected` has been used in the past. Avoiding this word going forward makes sure nobody carries old assumptions about what it means.
 1. I find `changed` to be more precise than `affected`. We have already seen that is not entirely clear whether `affected` is supposed to mean `involved` or `changed`.

HeikoKlare:
Fair point. To explain why we did not call it `changed` but `affected` is because next to changing a property of an element, there are also the change types of creating and deleting an element. Since that element is actually not `changed`, we agreed to subsume `changed`, `created` and `deleted` objects under `affected` objects.
Maybe `affected` is not the optimal term, but I am not sure that `changed` is it either.
Anyway, that would be a (really) huge change, because several classes, metamodels, language specifications and other files are affected. So if we agree that `affected` should be replaced (by any other term), we should have an issue for that to be solved as soon as someone has the time for that.

Writing these lines I realized that with this distinction, `changedResource` would actually be the proper term, even with the current `affected` terminology, because creations and deletions do actually not affect a resource. So I will at least perform that change.

HeikoKlare:
I hope that my recent changes and the submitted issue reflect the proposals. If this is the case and you think the PR is ready to merge, please approve :-)

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/description/VitruviusChange.xtend
```diff
@@ -64,11 +64,11 @@ interface VitruviusChange extends URIHaving {
 	def Iterable<String> getAffectedEObjectIds();
 
 	/**
-	 * Returns the unique {@link Resource} affected by this change, i.e., the resource containing the affected {@link EObject}s.
+	 * Returns the unique {@link Resource} changed by this change, i.e., the resource containing the affected {@link EObject}s.
```
jGleitz, line 67:
```suggestion
	 * Returns the unique {@link Resource} changed by this change, i.e., the resource containing the changed {@link EObject}s.
```

jGleitz (changes requested):

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/description/VitruviusChange.xtend
jGleitz: This distinction is important, because it explains why the resource is, in fact, unique.

jGleitz: I think it might also make sense to 

 a) document that `change.changedResource.URI == change.URI`
 b) either remove `change.URI` (because it is unnecessary) or change its implementation to `return changedResource.URI`

HeikoKlare: I have adapted the documentation with the clarification of `changed` in comparison to `affected` and a depiction of the relation between the change and the resource `URI`.
`change.URI` is, however, neither unnecessary nor equal to `changeResource.URI`, such that I left the implementation as it is. Currently, `changedResource` is calculated using `URI`, so there is at least no redundancy.

PR #334, opened 2020-12-11
Legacy Vitruv application test

HeikoKlare:
This PR introduces the `LegacyVitruvApplicationTest` providing an API comparable to the one of the old `VitruviusApplicationTest` to allows an easy temporary adaptation of our applications.
The class is to be removed as soon as the changes are adapted to the new `VitruvApplicationTest`.

### DIFF
--- a/bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/LegacyVitruvApplicationTest.xtend
+++ b/bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/LegacyVitruvApplicationTest.xtend
@@ -12,7 +12,7 @@ import org.eclipse.emf.common.notify.Notifier
 import tools.vitruv.framework.correspondence.CorrespondenceModel
 
 /** 
- * DO NOT USE THIS CALL!
+ * DO NOT USE THIS TEST CLASS! Use {@link VitruvApplicationTest} instead.
  * This is a temporary fallback class for our existing application tests until we have adapted them to the new
  * framework defined by the {@link VitruvApplicationTest}.
  * It serves as a temporary proxy to be removed after their adaptation and should not be used to implement any 
--- a/bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/LegacyVitruvApplicationTest.xtend
+++ b/bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/LegacyVitruvApplicationTest.xtend
@@ -1,0 +1,100 @@
+package tools.vitruv.testutils
+
+import tools.vitruv.testutils.VitruvApplicationTest
+import java.util.List
+import tools.vitruv.framework.change.description.PropagatedChange
+import org.eclipse.emf.ecore.resource.Resource
+import org.eclipse.emf.ecore.EObject
+import java.nio.file.Path
+import tools.vitruv.framework.change.description.VitruviusChangeFactory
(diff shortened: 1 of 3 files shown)


jGleitz:
What is the use case for `isResource` and `isNoResource` (as opposed to `exists` and `doesNotExist`?)

HeikoKlare:
They are defined for `URI` rather than `Resource`.

jGleitz:
Ah, I see. Can we improve the naming? While reading the code, I didn’t spot that there is a difference between `ResourceExistingMatcher` and `ResourceExistenceMatcher`. Maybe name the URI based `URIPoinstToResourceMatcher` or something else containing the name ‘`Uri`’?

HeikoKlare:
When trying to figure out why the matcher does not work for `URI`s, I had the same problem of not seeing the different between `ResourceExistingMatcher` and `ResourceExistenceMatcher` until I realized that the former is not used at all. Of course, we can rename the matcher for clarification. I was also not that convinced of the method names `isResource` and `isNoResource`, because it's actually not about being a `Resource` but about a model being persisted at the given `URI`, so maybe some naming like `isPersisted` and `isNotPersisted` would also be a good improvement?

jGleitz:
What is the use case for `ResourceExistingMatcher` in the first case? When I wanted to check whether a model exists, I always write `assertThat(resourceAt(....), exists)`. Is this not always sufficient?

HeikoKlare:
The difference is that the `ResourceExistingMatcher` does not require a resource to be loaded but only checks whether some persistence at that `URI` exists. To be honest, I find it a little confusing that `assertThat(resourceAt(....), exists)` can fail although `resourceAt(...)` returns some `Resource`. Maybe we should also rename `exists`, because it's also about the existence of a perisistence. Then we may also remove the `ResourceExistingMatcher` and accept that `exists` (maybe with some other name) loads a resource even if it is not necessary.

jGleitz:
You are right, naming it `existsOnTheFilesystem` would be more specific. In the case that we expect the resource to exist, the idiom `assertThat(resourceAt(....), exists)` checks en passant that the resource does not only exist (i.e. there is any file there) but that it can also be parsed (i.e. it is a valid persistence).

HeikoKlare:
Great, I think that we agree here. Just regarding the naming: `existsOnTheFilesystem` may be too specific, because, as far as I know, a URI must not necessarily point to a file in the file system, but can also be some database persistence or something else. Maybe we could just call it `isPersisted` or simply `isSaved` or something comparable? The only problem may be that it suggests that the given state of the resource is actually persisted/saved, but that is nothing that the matcher ensures (and is supposed to ensure).

jGleitz:
> The only problem may be that it suggests that the given state of the resource is actually persisted/saved.

Yes, that is my objection.

Maybe we should only have a matcher for URIs that is called `exists`, because I think that is self-explanatory. I didn’t want this, because it means we need to expose a method that hands out URIs for `Path`s and an overload of `resourceAt` that accepts an `URI`. Initially, I wanted to hide EMF’s `URI` type completely from the tests.

HeikoKlare:
I understand that goal and see its value. I think it is comparable to the idea of `VURI`s that we have in Vitruv as well (and they actually abstract from `URI`s  as it is a more general concept in comparison to `Path`). However, `URI` is more general than `Path` is and we need, for example, to load `pathmap` resources in tests, such that we need a `resourceAt` implementation that accepts `URI`s anyway. So I think your proposal of only having a matcher for `URI`s is fine.

PR #336, opened 2020-12-14
Improve logging

HeikoKlare:
Several logging and failure handling improvements


bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.java
```diff
@@ -398,7 +400,7 @@ public Void call() throws Exception {
                     resource.delete(null);
                     ResourceRepositoryImpl.this.modelInstances.remove(vuri);
                 } catch (final IOException e) {
-                    logger.info("Deletion of resource " + resource + " did not work. Reason: " + e);
+                    logger.error("Deletion of resource " + resource + " did not work. Reason: " + e);
```
jGleitz, line 403: Preferably:

```suggestion
                    logger.error("Deletion of resource " + resource + " did not work.", e);
```

to give the logging framework control over formatting the exception.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/helper/FileSystemHelper.java
```diff
@@ -306,8 +286,9 @@ private File createFolderInFolder(final File parentFolder, final String folderNa
     }
 
     private File createFolderIfNotExisting(final File folder) {
+        this.LOGGER.trace("Creating folder: " + folder);
         if (!folder.exists()) {
-            folder.mkdir();
+            checkState(folder.mkdir(), "Folder " + folder + " could not be created");
         }
```
jGleitz, line 292: More elegant (instead of the whole `if` block):

```
Files.createDirectories(folder.toPath())
```

You can leave out the `exists` check and don’t have to check the return value, because the `createDirectories` will throw an exception with the same information you are currently providing by hand.

HeikoKlare: Perfect, thanks! I will add that.

PR #340, opened 2020-12-15
Add method for project refresh and build

HeikoKlare:
- Adds a method for refreshing a project and incrementally building it
- Updates existing utils for issuing incremental project builds

### DIFF
--- a/bundles/framework/tools.vitruv.framework.ui.monitorededitor/src/tools/vitruv/framework/ui/monitorededitor/ProjectBuildUtils.java
+++ b/bundles/framework/tools.vitruv.framework.ui.monitorededitor/src/tools/vitruv/framework/ui/monitorededitor/ProjectBuildUtils.java
@@ -79,12 +79,17 @@ public final class ProjectBuildUtils {
         LOGGER.debug("Refresh and build project " + project.getName());
         try {
             project.refreshLocal(IResource.DEPTH_INFINITE, null);
-            project.build(IncrementalProjectBuilder.INCREMENTAL_BUILD, null);
         } catch (CoreException e) {
-            String message = "Could not refresh or build project " + project.getName();
+            String message = "Could not refresh project " + project.getName();
             LOGGER.error(message, e);
             throw new IllegalStateException(message, e);
         }
-
+        try {
+            project.build(IncrementalProjectBuilder.INCREMENTAL_BUILD, null);
+        } catch (CoreException e) {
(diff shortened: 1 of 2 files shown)


bundles/framework/tools.vitruv.framework.ui.monitorededitor/src/tools/vitruv/framework/ui/monitorededitor/ProjectBuildUtils.java
```diff
@@ -22,36 +24,67 @@ private ProjectBuildUtils() {
...
+        checkState(project.isOpen(), "Project must be open to be refreshed and built");
+        LOGGER.debug("Refresh and build project " + project.getName());
+        try {
+            project.refreshLocal(IResource.DEPTH_INFINITE, null);
+            project.build(IncrementalProjectBuilder.INCREMENTAL_BUILD, null);
+        } catch (CoreException e) {
+            String message = "Could not refresh or build project " + project.getName();
+            LOGGER.error(message, e);
+            throw new IllegalStateException(message, e);
+        }
```
jGleitz, line 87: If we had only the `refreshLocal` call in the `try` block and called our own `buildIncrementally` _outside_ the try block, the error and log messages could be more specific.

HeikoKlare: The `buildIncrementally` method is specific for a single builder. Running an incremental project build with all builders degrades to just calling `build` on the project. We could split refresh and build into two `try-catch` blocks to have more specific error and log messages?

jGleitz: > The buildIncrementally method is specific for a single builder. Running an incremental project build with all builders degrades to just calling build on the project. 

Ah, I didn’t catch that.

> We could split refresh and build into two try-catch blocks to have more specific error and log messages?

Would be a little nicer, to see which step actually went wrong. I don’t know whether that will be clear from the `CoreException`.

HeikoKlare: Yes, that makes sense. I have separated the two steps in a recent commit.

PR #357, opened 2020-12-22
Adapt build process to changed parent POM 

HeikoKlare:
- Updates to parent POM 0.9.0, which especially replaces most marker files for MWE2 workflows with conventions for MWE2 file placement and reconfigures the MWE2 workflows in a more reusable way
- Fixes versions of updatesite dependencies
- Corrects aggregated updatesite and fixes its version
- Removes unnecessary project-specific settings


HeikoKlare:
Drops the build time to less than 10 minutes 🎉

PR #364, opened 2021-01-08
Non-transactional Test View

HeikoKlare:
Introduces a non-transactional interface for starting and stopping recording, as well as propagating changes in tests.
* Introduces a specialization of `TestView` used by the `LegacyVitruvApplicationTest`
* Cleans up the interface of `VitruvApplicationTest`, especially by not exposing the `ChangeRecorder`
* Cleans up the `LegacyVitruvApplicationTest` and reduces it to the `NonTransactionalTestView` interface consisting of start and stopping recording, as well as propagating changes

Additionally, the PR ensures that resources of test views are only saved if they contain elements and are otherwise deleted, as the changes removed the contained elements.


bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/NonTransactionalChangePublishingTestView.xtend
```diff
@@ -0,0 +1,35 @@
+package tools.vitruv.testutils
+
+import java.nio.file.Path
+import tools.vitruv.framework.userinteraction.PredefinedInteractionResultProvider
+import org.eclipse.emf.common.notify.Notifier
+import tools.vitruv.framework.vsum.VirtualModel
+
+class NonTransactionalChangePublishingTestView extends ChangePublishingTestView implements NonTransactionalTestView {
```
jGleitz, line 8: Why do we need this class if `ChangePublishingTestView` already (effectively) implements `NonTransactionalTestView`?

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/NonTransactionalTestView.xtend
```diff
@@ -0,0 +1,48 @@
...
+import org.eclipse.emf.common.notify.Notifier
+import java.util.List
+import tools.vitruv.framework.change.description.PropagatedChange
+
+/**
+ * An extension of the {@link TestView}, which provides non-transactional options for change recording, i.e.,
+ * for manually starting and stopping recording changes of elements and for propagating all changes recorded
+ * since the last propagation.   
+ */
+interface NonTransactionalTestView extends TestView {
```
jGleitz, line 12: Does it make sense to mark this interface and all its methods as `@Deprecated`, with the rationale that using it will produce tests that are difficult to reason about?

Our do you see use cases where using it is legitimate (for non-historical use cases).

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/NonTransactionalChangePublishingTestView.xtend
HeikoKlare: The intention is that `ChangePublishingTestView` is only supposed to be a `TestView` but not a `NonTransactionalTestView`. If we make `ChangePublishingTestView` implement the `NonTransactionalTestView` interface, it would expose all the according methods. It does (effectively) implement that interface as it requires the according functionality anyway, but it is not supposed to expose it. In fact, the `NonTransactionalTestView` is expected to be removed as soon as all tests are refactored to to use the `VitruvApplicationTest`. I would expect it to be easier to remove the interface if a remaining class did not implement it, because otherwise someone may have accidentally depended on the additional provided interface.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/NonTransactionalTestView.xtend
HeikoKlare: I am torn between marking the `NonTransactionalTestView` as `@Deprecated`, like the `LegacyVitruvApplicationTest` was before, and just documentation that it should not be used for new implementations.
On the one hand, I do not see any future use cases for this test view and the `LegacyVitruvApplicationTest`, which is definitely an argument for marking it as `@Deprecated`.
On the other hand, I do not expect us to rework all the existing tests to use the `VitruvApplicationTest` within the next few months (although I hope that we will be able to do so). As long as it is unclear when we will be able to remove these classes (or whether this will be the case at all), I would not like to have warnings due to using deprecated methods within all the projects.

This is why I added an according documentation to the `LegacyVitruvApplicationTest` (and I hope that `Legacy` also makes it clear without reading the documentation), but, of course, using build-in language mechanisms would be even better.
I am still open for discussion, but just wanted to clarify the rationale for my decision.

jGleitz: I think `@Deprecated` is the right tool for the job. It will produce a warning whenever someone tries to use these types. This is more effective than documentation. Documentation is often not read, warnings annoy and are harder to miss.

True, having a lot of `@Deprecated` warnings going forward will also be annoying. But I think that is the lesser evil.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/NonTransactionalChangePublishingTestView.xtend
jGleitz: I disagree here: Removing `NonTransactionalTestView` should always be done by simplifying `ChangePublishingTestView`. So the argument about removing the interface does not hold water, from my point of view.

Additionally, nobody should reference `ChangePublishingTestView` test view in their code, but only `TestView`. Hence, `NonTransactionalTestView` will not really be visible to clients, since they should depend on the interface `TestView`, not its implementation.

If `NonTransactionalTestView` is additionally marked as `@Deprecated`, I am quite certain that nobody will depend on this interface if they can avoid it.

HeikoKlare: I am still not completely convinced, because `ChangePublishingTestView` will change its interface as soon as we remove `NonTransactionalTestView`, no matter whether someone actually depends on the implementation or only the to-be-removed interface, which would be avoided by having an additional class.
However, I see that the point that this will unlikely be the case and thus proposed an improvement (by removing the `NonTransactionalChangePublishingTestView` in a recent commit. Let me know whether the solution convinces you :-)

jGleitz: > `ChangePublishingTestView` will change its interface as soon as we remove `NonTransactionalTestView`

Yes, it will. However, I currently don’t see why this should be a problem.

> Let me know whether the solution convinces you :-)

Yes, I think this is the better solution.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/NonTransactionalTestView.xtend
HeikoKlare: I totall agree that `@Deprecated` is (technically) the right tool and that it is much more effective than documentation.
However, I can tell you the story that we will write half a year from now: A single person created a single new specialization of `VitruvApplicationTest`. If he or she actually ignored the documentation and the class qualifier and thus used the `LegacyVitruvApplicationTest`, at least some colleague has noticed the usage of `LegacyVitruvApplicationTest` and enforced a fix.
Meanwhile, three other persons modified the applications. All of them were annoyed by all the useless warnings, filtering warnings for deprecation also hides relevant warnings for using deprecated APIs. And at least one of them was too lazy or unable to properly filter his or her warnings and thus introduced several new warnings and thought that if so many warnings preexisted, some more warnings will not be a problem.

What I just want to point out: From my experience using `@Deprecated` in this case will introduce more problems than it will solve. The interfaces are not to be removed within a fixed period of time (like normally inteded when using `@Deprecated`), but will simply be removed as soon as we have adapted the applications.

jGleitz: I see your point. The fact that we realistically expect people to disable warnings instead of fixing the problems tells stories about the state this project is in and will be in. Hence, maybe it is futile to discuss the best approach. Chose the path you prefer, it seems that either way nothing will get cleaned up. I wash my hands of it.

HeikoKlare: > The fact that we realistically expect people to disable warnings instead of fixing the problems tells stories about the state this project is in and will be in.

> it seems that either way nothing will get cleaned up

No one said that problems won't get fixed. As you know, we cannot rewrite the whole bunch of tests within a few hours, but it will take some time. If you pay for an employee who does the job, fine :-) Otherwise, we will have to wait until someone has the time to do the job.
In addition, the core of the problem with our tests is not the usage of an interface for change recording that does not ensure transactionality. Sure, using that interface enforces a better structure, but in fact this is not the reason for most of the problems.

PR #390, opened 2021-01-19
Dispose Virtual Model before Closing a Test View

HeikoKlare:
Otherwise closing the view may try to delete files which are still loaded in the resource set of the virtual model, potentially leading to access problems.

### DIFF
--- a/bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/VitruvApplicationTest.xtend
+++ b/bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/VitruvApplicationTest.xtend
@@ -56,8 +56,8 @@ abstract class VitruvApplicationTest implements CorrespondenceModelContainer, Te
 
 	@AfterEach
 	def final package void closeAfterTest() {
+		virtualModel.dispose()
 		testView?.close()
-		virtualModel.dispose
 	}
 
 	override CorrespondenceModel getCorrespondenceModel() { virtualModel.correspondenceModel }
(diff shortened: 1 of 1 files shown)


bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/VitruvApplicationTest.xtend
```diff
@@ -56,8 +56,8 @@ abstract class VitruvApplicationTest implements CorrespondenceModelContainer, Te
 
 	@AfterEach
 	def final package void closeAfterTest() {
+		virtualModel.dispose()
```
jGleitz, line 59:
```suggestion
		virtualModel?.dispose()
```

I think we should make this call null-safe as well, because creating the `VirtualModelConfiguration` can already fail. The null-safe call avoids unncessary suppressed `NullPointerException`s

HeikoKlare: Maybe we should do some check already during setup? If the `VirtualModel` is not created properly, it will fail on first use of the `VirtualModel` and not when closing the test.

jGleitz: No, if the `virtualModel` cannot be created properly, no test will be executed (because the `@BeforeEach` method threw an exception). However, `@AfterEach` methods will still be executed. In such a case, the `virtualModel` might be null, which adds another suppressed `NullPointerException` to the exception from the  `@BeforeEach` method .

HeikoKlare: Ah, I see. Thanks! Then it absolutely makes sense.

PR #411, opened 2021-02-09
Remove Duplicate Adapter Removal

HeikoKlare:
Disposing the `ResourceSetImpl` unnecessarily cleared the adapters of each object, although `unload` does that anyway.
Under specific conditions, this can lead to concurrent modifications of the resource list and thus a `ConcurrentModificationExeception` in the iterator.
This PR simply removes the duplicate adapter removal.


bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.xtend
```diff
@@ -337,13 +337,11 @@ class ResourceRepositoryImpl implements ModelRepository, CorrespondenceProviding
 	override Resource getModelResource(VURI vuri) {
 		getModelInstanceOriginal(vuri).resource
 	}
-	
+
 	def dispose() {
 		resourceSet.transactionalEditingDomain?.dispose
-		resourceSet.resources.forEach[
-			allContents.forEach[eAdapters.clear]
```
jGleitz, line 344: It’s great that this line is removed, because it could cause `StackOverflowErrors` because it resolves proxies (the fact that it resolves proxies for something we are going to unload was bad in and of itself).

PR #413, opened 2021-02-09
Close OutputStream in DomainAwareResourceSetTests

HeikoKlare:
Avoids file access exceptions on Windows systems due to unclosed `OutputStream`s in tests.

### DIFF
--- a/tests/framework/tools.vitruv.framework.domains.tests/src/tools/vitruv/framework/tests/domains/repository/DomainAwareResourceSetTest.xtend
+++ b/tests/framework/tools.vitruv.framework.domains.tests/src/tools/vitruv/framework/tests/domains/repository/DomainAwareResourceSetTest.xtend
@@ -38,7 +38,7 @@ class DomainAwareResourceSetTest {
 		resourceFactoryRegistry = new TestResource.Factory.Registry
 		awareOfDomains(TestDomain.repository)
 	]
-	
+
 	@BeforeAll
 	def static void writeResourceToLoad(@TestProject Path testProject) {
 		aetResourceUri = URI.createFileURI(testProject.resolve("example".allElementTypes).toString)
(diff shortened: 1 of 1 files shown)


jGleitz:
Ouch. Technically, we should make sure that the streams are closed even if an exception is thrown

HeikoKlare:
> Technically, we should make sure that the streams are closed even if an exception is thrown

Of course, they should. I have proposed an improved solution. 

Sorry for the formatting changes. There seems to be an imcompatibility between the standard formatters we use?

jGleitz:
> There seems to be an imcompatibility between the standard formatters we use 

I use none. That’s whey I have all that unnecessary whitespace. All formatters I tried would not only produce, but _enforce_ horrible, unreadable results for chained calls and/or lambdas. So I do formatting by hand.

tests/framework/tools.vitruv.framework.domains.tests/src/tools/vitruv/framework/tests/domains/repository/DomainAwareResourceSetTest.xtend
```diff
@@ -88,92 +88,94 @@ class DomainAwareResourceSetTest {
...
 		]
 		assertThat(TestResource.lastSaveOptions, is(<Object, Object>emptyMap))
-		
-		val outputStream = domainAwareResourceSet.URIConverter.createOutputStream(aetResourceUri)
-		domainAwareResourceSet.createResource(aet2ResourceUri) => [
-			contents += aet2.Root2
-			save(outputStream, emptyMap)
-		]
+
+		try(val outputStream = domainAwareResourceSet.URIConverter.createOutputStream(aetResourceUri)) {
```
jGleitz, line 92: Wait, since when can we use try-with-resources in Xtend? And I’ve been using that damned `@CloseResource` annotation all the time!

HeikoKlare: I did not think about that. Seems like it was added in 2.18 which was released in 2019.

HeikoKlare:
>I use none. That’s whey I have all that unnecessary whitespace. All formatters I tried would not only produce, but enforce horrible, unreadable results for chained calls and/or lambdas. So I do formatting by hand.

Unfortunately, that's true. It is sometimes possible to enforce proper formatting by adding empty comments, but that's neither a good nor a universal solution. Maybe we could try to apply standard formatting whenever there is nothing within a file that is broken by the formatter?

tests/framework/tools.vitruv.framework.domains.tests/src/tools/vitruv/framework/tests/domains/repository/DomainAwareResourceSetTest.xtend
jGleitz: Very good to know!

jGleitz:
>  Maybe we could try to apply standard formatting whenever there is nothing within a file that is broken by the formatter?

I can try to remember to auto-format such files. But I will surely not enable format on save again.

HeikoKlare:
> But I will surely not enable format on save again.

Of course. I do not have it enabled either, for the same reasons.

PR #420, opened 2021-02-17
Add Missing Dependencies to Testutils Feature

HeikoKlare:
Add dependencies for testutils feature necessary since #415.

### DIFF
--- a/features/tools.vitruv.testutils.feature/feature.xml
+++ b/features/tools.vitruv.testutils.feature/feature.xml
@@ -33,6 +33,10 @@
       <import plugin="org.eclipse.emf.compare.rcp"/>
       <import plugin="org.junit.jupiter.params"/>
       <import plugin="org.eclipse.xtext.xbase.lib"/>
+      <import plugin="org.eclipse.xtext.ui.testing"/>
+      <import plugin="org.slf4j.api"/>
+      <import plugin="ch.qos.logback.classic"/>
+      <import plugin="ch.qos.logback.core"/>
    </requires>
 
    <plugin


features/tools.vitruv.testutils.feature/feature.xml
```diff
@@ -33,6 +33,10 @@
       <import plugin="org.eclipse.emf.compare.rcp"/>
       <import plugin="org.junit.jupiter.params"/>
       <import plugin="org.eclipse.xtext.xbase.lib"/>
+      <import plugin="org.eclipse.xtext.ui.testing"/>
+      <import plugin="org.slf4j.api"/>
+      <import plugin="ch.qos.logback.classic"/>
+      <import plugin="ch.qos.logback.core"/>
```
jGleitz, line 39: The two logback dependencies should explicitly *not* be necessary! I wrote the code in a way that we can make sure that they are _not_ forced onto clients.

jGleitz: It was probably a mistake that I declared them in the `MANIFEST.MF`, though.

jGleitz (changes requested):
We should not force Logback onto clients

features/tools.vitruv.testutils.feature/feature.xml
HeikoKlare: Hm, we could add `resolution:=optional` to the `MANIFEST.MF` to indicate that these dependencies are not mandatory. The drawback is that setting up Vitruv gets more complicated, because you need to explicitly install these dependencies into your Eclipse. If you only use the `testutils` project this is not a problem, but if you want to import it into you workspace and modify it, you will have to resolve the dependencies on your own.

HeikoKlare: And yes, having them declared in the `MANIFEST.MF` like we have now will install them on clients whenever the `testutils` plugin is installed (even though they are not declared in the feature).

jGleitz: > if you want to import it into you workspace and modify it, you will have to resolve the dependencies on your own.

I see that this could be a problem in theory, but in practice, the dependencies should be pulled into the workspace through transitive dependencies anyway.

> And yes, having them declared in the MANIFEST.MF like we have now will install them on clients whenever the testutils plugin is installed (even though they are not declared in the feature).

What happens if we declare them as optional imported packages? I’d expect that this would not pull any additional bundles in (since we don’t specify which).

```
Import-Package: org.slf4j;resolution:=optional,
 ch.qos.logback.classic;resolution:=optional,
 ch.qos.logback.classic.spi;resolution:=optional,
 ch.qos.logback.classic.encoder;resolution:=optional,
 ch.qos.logback.core;resolution:=optional
```

jGleitz: See https://github.com/vitruv-tools/Vitruv/pull/422

HeikoKlare: >I see that this could be a problem in theory, but in practice, the dependencies should be pulled into the workspace through transitive dependencies anyway.

Unfortunately, I do not understand that comment.

> What happens if we declare them as optional imported packages? I’d expect that this would not pull any additional bundles in (since we don’t specify which).

I guess that the build will fail because it does not know where to take the packages from (and #422 seems to confirm that).

jGleitz: > Unfortunately, I do not understand that comment.

You claimed that if somebody wanted to modify the `testutils` project, they would need to install the logback dependencies by hand. And my counterargument is that the logback dependencies are likely to be already present because they will have been pulled in transitively by another bundle.

HeikoKlare: > they will have been pulled in transitively by another bundle.

I don't understand why that should be the case.

jGleitz: Well, it’s the case in my workspace :shrug:

jGleitz: My understanding is that as soon as some other bundle requires the logback dependencies (which some do), the dependencies are available to Eclipse and can be resolved.

HeikoKlare: In my Eclipse they are not available and I do actually not understand why we should hope that there are some _other_ bundles that require them.

jGleitz: For the same reason that the logback code is there in the first place: When running tests in Eclipse, some bundle(s) use(s) logback to log. I added code to mute those messages. If there weren’t any bundles using logback, we wouldn’t need the code. That’s why I expected that anybody would have the bundles in their target platform.

HeikoKlare: There is no transitive dependency from our code to logback. If you set up a clean (Modeling Tools) Eclipse and install Vitruv into it, logback is not installed.

jGleitz: >  If you set up a clean (Modeling Tools) Eclipse and install Vitruv into it, logback is not installed.

Yeah, but that’s not an issue, is it? Because in this situation, there is no reason why you should have logback around. It’s only an issue if you are a developer for Vitruv.

jGleitz: > There is no transitive dependency from our code to logback.

Yes, this is the reason why I don’t want to add one.

HeikoKlare:
Closed in favor of #430.

PR #424, opened 2021-02-18
UuidGenerator URI Resolution Fix

### LINKED ISSUE #423: Erroneous URI Calculation in UuidGeneratorAndResolverImpl
The changed URI calculation in `UuidGeneratorAndResolverImpl` in the `getResolvableUri` method to support resources with XMI IDs is erroneous:
https://github.com/vitruv-tools/Vitruv/blob/a45e46984b5c94d2609adf634ea9866749808fda/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend#L342-L347

The implementation assumes that the fragment delivered by `EcoreUtil.getRelativeURIFragmentPath` is relative to the resource of the elements, but this is unfortunately wrong.
`EcoreUtil.getRelativeURIFragmentPath` delivers the fragment relative to the topmost element in the containment hierarchy, whereas `eResource` may deliver a resource with some element within the containment hierarchy as its root element.

As a concrete example where this fails: A `Class` in the Java model delivers the resource of the `CompilationUnit` it is contained in. In the containment hierarchy, however, the `Package` containing the `CompilationUnit` is (or can be) the topmost element.
The resolver then tries to resolve a `compilationUnits` reference within a `CompilationUnit`, which obviously fails.
This is the reason for several failures of the kind:
`java.lang.IllegalArgumentException: The feature 'compilationUnits' is not a valid feature`

HeikoKlare:
Corrects the fragment-based URI calculation in the `UuidGeneratorAndResolverImpl` to support the case that the root container is not the root element of the resource of an object (see #423).

Fixes #423.


HeikoKlare:
✔️ Works with applications (tests)

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -340,11 +340,12 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 	
 	def private static getResolvableUri(EObject object) {
 		val resourceUri = object.eResource.URI
+		val resourceRoot = object.eResource.firstRootEObject
 		// we cannot simply use EcoreUtil#getURI, because object’s domain might use XMI	UUIDs. Since
 		// XMI UUIDs can be different for different resource sets, we cannot use URIs with XMI UUIDs to identify objects
 		// across resource sets. Hence, we force hierarchical URIs. This assumes that the resolved object’s graph
 		// has the same topology in the resolving resource set. This assumption holds when we use this method.  
-		val fragmentPath = EcoreUtil.getRelativeURIFragmentPath(null, object)
+		val fragmentPath = EcoreUtil.getRelativeURIFragmentPath(resourceRoot, object)
```
jGleitz, line 348: What happens if a resource has multiple root elements? In this case we’d need to find the one that contains the `object`, wouldn’t we?

jGleitz (changes requested):
I think we should start to be a lot more consequential with unit testing in Vitruv. So I’d suggest that we add a unit test for this particular bug.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: I thought about that as well. In general, Ecore is tied to using only a single root element and when having multiple root elements the first one is kind of "the central one".
Apart from transient states (like code generation for Xtext resources), the only situation I know in which multiple root elements are used are UML stereotypes. But, of course, we need to ensure that assigning UUIDs to stereotypes (or potentially other elements contained in further root elements) works properly. This will require some kind of search. I'll propose an according improvement.

jGleitz: It might make sense to first write a unit test for this situation, to be sure that it would really lead to a bug.

jGleitz: `getRelativeURIFragmentPath` contains this part:

```
if (internalEObject != ancestorEObject && ancestorEObject != null)
    {
      throw new IllegalArgumentException("The ancestor '" + ancestorEObject + "' is not an ancestor of '" + descendantEObject + "'");
    }
```

So we might also just not fix this situation, because we know we would see an exception should the situation ever arise (which is much better than having a silent failure).

HeikoKlare:
>I think we should start to be a lot more consequential with unit testing in Vitruv. So I’d suggest that we add a unit test for this particular bug.

Yes, I am still working on some tests to provide.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: I've added an according fix together with unit tests that led to a bug before.
In fact, the used fragment-based URIs were only a shortcut for resources with a single root elements (this is why the fragment is separated by `//`, as the number of the root element is contained in between (which is optional if there is only one).

HeikoKlare:
I've added some first test cases specifically for the mentioned bug as well as for the support of multiple root elements. 
They should be sufficient as regression tests for the problems I faced before.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -339,16 +339,31 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
+		val resource = object.eResource
+		val resourceRoot = if (resource.contents.size <= 1) {
+			object.eResource.firstRootEObject
+		} else {
+			// move up containment hierarchy until some container is one of the resource's root elements
+			var container = object
+			while (!resource.contents.contains(container) && container !== null) {
+				container = container.eContainer
+			}
+			checkState(container !== null, '''some container of «object» must be a root element of its resource''')
```
jGleitz, line 351:
```suggestion
			checkState(container !== null, "some container of %s must be a root element of its resource", object)
```

Avoids creating a `StringConcatenationClient` and building the result `String` on every invocation. Builds the string only if the condition fails.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -339,16 +339,31 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
+			container
+		}
 		// we cannot simply use EcoreUtil#getURI, because object’s domain might use XMI	UUIDs. Since
 		// XMI UUIDs can be different for different resource sets, we cannot use URIs with XMI UUIDs to identify objects
 		// across resource sets. Hence, we force hierarchical URIs. This assumes that the resolved object’s graph
 		// has the same topology in the resolving resource set. This assumption holds when we use this method.  
-		val fragmentPath = EcoreUtil.getRelativeURIFragmentPath(null, object)
+		val fragmentPath = EcoreUtil.getRelativeURIFragmentPath(resourceRoot, object)
+		val rootElementIndex = resource.contents.indexOf(resourceRoot)
```
jGleitz, line 359: We could remember this index from above (in the first case, it’s `0`, in the second case, we can replace the `contains` with `indexOf`)

tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend
```diff
@@ -0,0 +1,147 @@
...
+
+@ExtendWith(TestProjectManager, RegisterMetamodelsInStandalone)
+class UuidGeneratorAndResolverImplTest {
+	val ResourceSet resourceSet
+	val UuidGeneratorAndResolver uuidGeneratorAndResolver
+	var Path testProjectPath
+
+	new() {
+		resourceSet = new ResourceSetImpl().withGlobalFactories()
+		uuidGeneratorAndResolver = new UuidGeneratorAndResolverImpl(resourceSet, true)
```
jGleitz, line 31: I think the code is easier to read if the field initialization is in the same line as the field declaration (plus, you can omit types then).

But that is not something I’d insist on.

bundles/testutils/tools.vitruv.testutils.metamodels/metamodels/allElementTypesWithContainmentProxies.ecore
```diff
@@ -0,0 +1,52 @@
...
+  <eClassifiers xsi:type="ecore:EClass" name="NonRootObjectContainerHelperWithContainmentProxies"
+      eSuperTypes="#//IdentifiedWithContainmentProxies">
+    <eStructuralFeatures xsi:type="ecore:EReference" name="nonRootObjectsContainment"
+        upperBound="-1" eType="#//NonRootWithContainmentProxies" containment="true"/>
+  </eClassifiers>
+  <eClassifiers xsi:type="ecore:EClass" name="IdentifiedWithContainmentProxies" abstract="true">
+    <eStructuralFeatures xsi:type="ecore:EAttribute" name="id" lowerBound="1" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"
+        iD="true"/>
+  </eClassifiers>
+</ecore:EPackage>
```
jGleitz, line 52: Can you explain why we need an additional test model here?

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -339,16 +339,31 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
+			while (!resource.contents.contains(container) && container !== null) {
+				container = container.eContainer
+			}
+			checkState(container !== null, '''some container of «object» must be a root element of its resource''')
+			container
+		}
 		// we cannot simply use EcoreUtil#getURI, because object’s domain might use XMI	UUIDs. Since
 		// XMI UUIDs can be different for different resource sets, we cannot use URIs with XMI UUIDs to identify objects
 		// across resource sets. Hence, we force hierarchical URIs. This assumes that the resolved object’s graph
 		// has the same topology in the resolving resource set. This assumption holds when we use this method.
```
jGleitz, line 357: I think this comment should go to the top of the method, because it explains what we are planning to do.

tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend
```diff
@@ -0,0 +1,147 @@
...
+			save(null)
+		]
+		childResourceSet.createResource(URI.createFileURI(testProjectPath.resolve("nonRoot.aet").toString)) => [
+			contents += nonRoot
+			save(null)
+		]
+
+		// UUID generation and registration
+		val generatedUuid = childUuidGeneratorAndResolver.generateUuid(nonRoot)
+		val parentResolverNonRoot = resourceSet.getEObject(EcoreUtil.getURI(nonRoot), true)
```
jGleitz, line 61: Could read a little nicer if we added a `static extension` import for `EcoreUtil.getURI` and wrote `nonRoot.URI`. Once again, that’s a nit-pick, I am not going to insist on it.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -339,16 +339,31 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
-		val resourceUri = object.eResource.URI
+		val resource = object.eResource
+		val resourceRoot = if (resource.contents.size <= 1) {
+			object.eResource.firstRootEObject
+		} else {
+			// move up containment hierarchy until some container is one of the resource's root elements
+			var container = object
+			while (!resource.contents.contains(container) && container !== null) {
+				container = container.eContainer
+			}
```
jGleitz, line 350: `EcoreUtil` always does a check for cycles in the containment hierarchy when traversing the tree. Should we do this as well?

If we do, we might consider whether we should call `getRelativeURIFragmentPath` at all. It basically does the same traversal of the tree upwards (but with a check for cycles). So we could also just build the URI fragment ourselves with `StringBuilder.insert` while traversing upwards.

jGleitz (changes requested):
I have some remarks. Most of them are questions.

tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend
HeikoKlare: That's a great proposal! I'll improve that.

tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend
HeikoKlare: I have to admit that I was not sure whether the order of field initalization is defined and deterministic or whether `resourceSet` argument can be `null`. I forgot to check that but will have a look at it.

bundles/testutils/tools.vitruv.testutils.metamodels/metamodels/allElementTypesWithContainmentProxies.ecore
HeikoKlare: Actually, we do not need an additional model, but a different type of generated code. As the name indicates, the model has `containmentProxies` enabled, which is the think that enables Java elements to be contained in a `Resource` and in another container at the same time (which is not possible otherwise).
We could reuse the `allElementTypes` metamodel and just define an additional genmodel, but I found it bad to have the same classes multiple times just placed in the different packages (i.e., having `Root`, `NotRoot` etc. duplicated).

jGleitz: Does enabeling `containmentProxies` break anything else? I’d expect that as long as you do not build weird containment hierarchies, code generated with the option behaves just like code generated without the option.

If so, we could avoid having an additional test metamodel.

jGleitz: Note to self: We need to change `ChangeRecorder`, because containment proxies break one of its assumptions: https://github.com/vitruv-tools/Vitruv/blob/master/bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/ChangeRecorder.xtend#L268-L269

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -340,13 +340,16 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
 		val resourceRoot = if (resource.contents.size <= 1) {
 			object.eResource.firstRootEObject
 		} else {
 			// move up containment hierarchy until some container is one of the resource's root elements
 			var container = object
-			while (!resource.contents.contains(container) && container !== null) {
+			rootElementIndex = resource.contents.indexOf(container)
+			while (rootElementIndex == -1 && container !== null) {
 				container = container.eContainer
+				rootElementIndex = resource.contents.indexOf(container)
```
jGleitz, line 352: `container` might be `null` here!

jGleitz: The whole loop might most easily be written like this. However, it contains a nasty side-effect:

```Xtend
var container = object
while (container !== null && (rootElementIndex = resource.contents.indexOf(container)) != -1) {
    container = container.eContainer
}

File: bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend (hunk 8 above)
Comment (HeikoKlare) [reply bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend:353]: That's an interesting point. I assumed that there cannot be cycles in the containment hierarchy. Actually, the `eResource` implementation simply traverses the containments without a check for cycles. So as far as I understand we will already run into a problem when trying to get the resource, do I think we can ignore that case?

Independent from the previous point, I would not replace the usage of `getRelativeURIFragmentPath` with an own implementation using a `StringBuilder` to stay compatible with potential changes in the implementation and avoid duplication of that logic.

Comment (jGleitz) [reply bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend:353]: > I assumed that there cannot be cycles in the containment hierarchy

Me too. Then I saw that `EcoreUtil` was check for it.

That’s untrue. From `BasicEObjectImpl#eInternalResource`:

```Java
// Since the cycle is detected by checking if we hit "this" again, after many iterations we'll call this method recursively 
// in case we started with something that wasn't part of a cycle but later traversed up to a cycle.
//
if (++count > 100000)
  {
    return eContainer.eInternalResource();
   }
    result = eContainer.eDirectResource();
    if (result != null || eContainer == this)
    {
      break;
    }
```

jGleitz: To be clear: I am not saying that we should check for cycles, I just want to bring up that EMF seems to be prepared for that kind of problem.

bundles/testutils/tools.vitruv.testutils.metamodels/metamodels/allElementTypesWithContainmentProxies.ecore
HeikoKlare: >I’d expect that as long as you do not build weird containment hierarchies, code generated with the option behaves just like code generated without the option.

>If so, we could avoid having an additional test metamodel.

I though about that as well, because I also didn't want to introduce a further, almost redundant metamodel. However, my intent was the following: Every metamodel that we use (except for the JaMoPP metamodel) has `containmentProxies` disabled and this is also the default setting when creating a genmodel. Thus, even if it only makes (or should make) a difference under rather unusual circumstances, I wanted to keep the metamodel used through all of our basic tests conforming to the standard configuration and thus introduced an additional one for this specific case.

However, I am open to discuss this, because having fewer almost redundant metamodels would be great. I just wanted to clarify the reasons for my decision.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: Ah, that's right. That kind of cycle check is "interesting". Maybe it's too late for me, but then `eResource` should return `null` in case there is a cycle, shouldn't it? So our implementation will still fail, as it implicitly assumes `eResource` not to return `null`. But in fact, this cannot happen as the caller checks that the object has a resource. So I guess, we cannot run into the situation that there is such a cycle?

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: Yes, that's probably most elegant, even with the side effect.
Depending on the average structure/size of models it could even be beneficial to copy the resource contents into a `HashSet` and make the containment traversal by checking containment in that set. Then we could, like in the initial solution, calculate the index afterwards. But I am not sure whether the performance and memory overhead for setting up the set outweighs the performance improvement during the search.

HeikoKlare: At least, the loop header has to check for `==` instead of `!=` ;-)

jGleitz: > But I am not sure whether the performance and memory overhead for setting up the set outweighs the performance improvement during the search.

I doubt it.

HeikoKlare: Me too. I've already implemented your proposal, but tests are still running.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: Fixed

bundles/testutils/tools.vitruv.testutils.metamodels/metamodels/allElementTypesWithContainmentProxies.ecore
jGleitz: I think wanting to have the example models to match the ‘typical’ case is sensible. However, I think in this instance it’s quite unlikely that there can be a bug that we would not catch with `containmentProxies` enabled, but would catch with `containmentProxies` disabled. And even if there was, we could test in with all element types 2 once we found it.

Because of this, I think having the one less example metamodel trumps the concern for realistic examples in this instance.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
jGleitz: It seems you’re right. Great, that means we’re good!

bundles/testutils/tools.vitruv.testutils.metamodels/metamodels/allElementTypesWithContainmentProxies.ecore
HeikoKlare: That's right. My concern was that `containmentProxies` are a slight derivation of the typical behavior, which many developers will potentially not be aware of that it exists. If it actually leads to some bug, it may be cumbersome to find it (like it was for me to figure out why Java models behave differently, as I was not aware of it either).
But you are probably right that it's pretty unlikely that it will lead to problems at all, especially because the AllElementTypes metamodel will unlikely be used with proxy containments.
So I've removed that additional metamodel.

jGleitz (approved):
Looks good to me!

HeikoKlare:
Perfect, thanks for all the great feedback!
I've successfully validated the changes with the applications tests again, so I'll merge this PR.

Issue #426, opened 2021-02-19
UUID Issue in Addresses-Recipients Demo Tests


HeikoKlare:
In the Addresses-Recipients tests in `tools.vitruv.applications.demo.addressesrecipients.tests`, a bunch of warnings occurs because UUIDs for elements with proxy URIs are not found. This behavior seems to occur since #419, in which the change recorders were unified.
To me, this looks like a bug revealed by the changes in #419, but we should figure out the reason and fix it. The warnings especially slow down the build by several minutes due to thousands of warnings being printed.

jGleitz:
These warnings occur only when executing multiple test methods, never when executing only one. This may point to an issue with the test setup.

jGleitz:
Yes, the root issue here is the code that is generated by the Mappings Language: Every mapping has an `AbstractMappingRuntime`, which is a singleton. My understanding is that this seems to be something like a poor man’s intermediate model. However, since it is a singleton, it is not tied to the lifecycle of a VSUM (even though it absolutely should).

I could reproduce the problem in the debugger: After a test has been executed, the`AbstractMappingRuntime#instance` (e.g. `DeleteAdRootXReRootMappingInstancesRoutine#instance`) still references an object from the _last_ test run (which is a proxy by now). It is completely valid that the `UUIDGeneratorAndResolverImpl` doesn’t find a UUID for this object.

I _think_ that the mappings continue to work because the `InternalCorrespondenceModelImpl` doesn’t return correspondences for elements without a UUID.

HeikoKlare:
Thank you for the investigation! Taking a look into that project shows that it actually has nothing to do with the current Mappings language implementation :-/
It was used as a development project, in which Mappings were defined by hand to develop the language against. But after the language was implemented, the scenario was not re-implemented with that language but still uses some old development artifacts (such as the `AbstractMappingRuntime`).

In consequence, the project only serves as a documentation of some design ideas for the Mappings language and how that language could have been implemented (in addition to a really minimalist example case study).
I am not sure whether the project gives us a single benefits and thus whether we should keep the project at all. Adapting a pretty broken implementation (`AbstractMappingRuntime`) that no one needs anyway does not make that much sense.

PR #428, opened 2021-02-19
Remove TUIDs

### LINKED ISSUE #55: Record element modification by Reactions for TUID update
Reactions perform modifications on target models that are not reflected in any way. They are not used for transitively propagating them by triggering further Reactions and are also not investigated for TUID updates.

TUID updates for elements modified by Reactions are currently performed by the Reactions code that triggers the update procedure for certain elements at the correct points of time. This has the drawback that modifications on the Reactions language have to be performed correctly so that the TUID update is still performed correctly. This is error-prone and elaborate.
We should record changes in the models of the VSUM during change propagation through Reactions and perform just-in-time TUID updates just like during modifications by the user. This will potentially require us to perform resource changes just-in-time as the change recorder has to be started for newly created models.

### LINKED ISSUE #61: TUID resolution ignores reference name
Some time ago, we added the reference name to the TUID to make the TUID more unambiguous and improve resolution.
While the TUID calculator adds the reference name to the TUID (at least in cases in which the attribute is set), the resolver does not use the reference name for restricting the lookup. In fact, it ignores it and thus potentially procudes false results. 

We have to update the TUID resolver to use the reference name for retrieving the matching element. Besides, this also improves resolution performance. Furthermore, we have to update the TUID calculation to also add the reference name when the attribute is not set.

### LINKED ISSUE #114: TUID management with several virtual models
The `TuidManager` is **not** properly working if there are more than one `VirtualModel`. For normal tests this is adequate but for the versioning mechanism there have to be more `VirtualModel`s with different state to compare them to each other. Because of the registration of the `CorrespondenceModel`s at the TuidManager all Tuids in all `VirtualModel`s will be updated, not only the Tuids in the current `VirtualModel` under observation.

### LINKED ISSUE #156: UUIDs in correspondence model
To get rid of TUIDs and problems related to them, the next step is to remove them from the correspondence model. Since every element now has a UUID, on which the complete `VirtualModel` and all change application mechanisms depend, we can also use that mechanism for correspondences.

HeikoKlare:
This PR removes TUIDs and completely moves to UUIDs.
- Remove `tools.vitruv.framework.tuid` project
- Remove TUIDs from domains, especially removing `TuidAwareVitruvDomain` and all other domain-specific logic related to TUIDs. Cleans up the `AbstractVitruvDomain` implementation, which also contained further logic only necessary for TUIDs.
- Remove TUIDs from correspondences and the correspondence model. Especially the `InternalCorrespondenceModelImpl` contained a bunch of functionality for TUID handling, which is completely removed.
- Remove obsolete command execution from `InternalCorrespondenceModelImpl`, which bloated up the code.
- Remove multiple datatsypes from `tools.vitruv.framework.util`, which were only used by TUID mechanism. Additionally, move tests from this project to a separate test projects and cleanup the dependencies.

Fixes #55 (or makes it obsolete).
Fixes #61 (or makes it obsolete).
Fixes #114 (or makes it obsolete).
Fixes #156.


bundles/framework/tools.vitruv.framework.correspondence/src/tools/vitruv/framework/correspondence/impl/InternalCorrespondenceModelImpl.xtend
```diff
@@ -253,240 +55,194 @@ class InternalCorrespondenceModelImpl extends ModelInstance implements InternalC
...
-				new HashSet<Correspondence>()
-			oldTuidList2Correspondences.add(
-				new Pair<List<Tuid>, Set<Correspondence>>(oldTuidList, correspondencesForOldTuidList))
+	}
+	
+	private def haveUuids(List<EObject> eObjects) {
+		if (eObjects.forall[uuidResolver.hasPotentiallyCachedUuid(it)]) {
+			return true
+		} else {
+			logger.warn("UUID resolver has no UUID for one of the elements: " + eObjects)
```
jGleitz, line 155: Can we make this an exception now? As far as I can see, this is an erroneous state that point to a bug and continuing makes no sense.

tests/framework/tools.vitruv.framework.vsum.tests/src/tools/vitruv/framework/tests/vsum/VsumTest.java
```diff
@@ -49,11 +48,11 @@ void acquireTestProjectFolder(@TestProject final Path testProjectFolder) {
...
 
-	private static final VitruvDomain UmlDomain = new ConcreteTuidAwareVitruvDomain("UML", Uml_mockupPackage.eINSTANCE,
-		new AttributeTuidCalculatorAndResolver(Uml_mockupPackage.eINSTANCE.getNsURI(), "id"), UML_FILE_EXT);
+	private static final VitruvDomain UmlDomain = new AbstractVitruvDomain("UML", Uml_mockupPackage.eINSTANCE,
+		UML_FILE_EXT) {};
 
-	private static final VitruvDomain PcmDomain = new ConcreteTuidAwareVitruvDomain("PCM", Pcm_mockupPackage.eINSTANCE,
-		new AttributeTuidCalculatorAndResolver(Pcm_mockupPackage.eINSTANCE.getNsURI(), "id"), PCM_FILE_EXT);
+	private static final VitruvDomain PcmDomain = new AbstractVitruvDomain("PCM", Pcm_mockupPackage.eINSTANCE,
+		PCM_FILE_EXT) {};
```
jGleitz, line 55: Why do those exist in the first place? We have domains in the testutils domains bundle.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/TestLogging.xtend
```diff
@@ -56,8 +54,6 @@ class TestLogging implements BeforeAllCallback {
 		rootLogger.level = ERROR
 		VITRUV_LOG_ROOTS.forEach [logger.level = toLevel(desiredLogLevel, WARN)]
 		if (System.getProperty(VM_ARGUMENT_ENABLE_ID_LOGGERS) != "true") {
-			TuidManager.logger.level = ERROR
-			TuidCalculatorAndResolverBase.logger.level = OFF
 			UuidGeneratorAndResolverImpl.logger.level = ERROR
```
jGleitz, line 57: Can you try whether removing this line is already feasible? (The last time I tried, the number of messages was overwhelming).I think in the long run, we should set the level of `UuidGeneratorAndResolverImpl` to `WARN` as well, so that we can find bugs.

bundles/framework/tools.vitruv.framework.correspondence/src/tools/vitruv/framework/correspondence/impl/InternalCorrespondenceModelImpl.xtend
```diff
@@ -253,240 +55,194 @@ class InternalCorrespondenceModelImpl extends ModelInstance implements InternalC
...
+	private def Set<Correspondence> getCorrespondencesForUuids(List<String> uuids) {
+		return this.correspondences.correspondences.filter[AUuids.containsAll(uuids) || BUuids.containsAll(uuids)].toSet
 	}
 
-	private def EObject resolveEObjectFromTuid(Tuid tuid) {
-		tuidResolver.resolveEObjectFromTuid(tuid);
+	override <C extends Correspondence> Set<List<EObject>> getCorrespondingEObjects(Class<C> correspondenceType,
+		Predicate<C> correspondencesFilter, List<EObject> eObjects, String tag) {
+		return getCorrespondences(correspondenceType, correspondencesFilter, eObjects, tag)
+			.mapFixed[resolveCorrespondingObjects(eObjects)].toSet
```
jGleitz, line 180: This first creates a list, just to convert it into a set directly afterwards. Better:

```suggestion
			.mapFixedTo(new LinkedHashSet) [resolveCorrespondingObjects(eObjects)]
```

bundles/framework/tools.vitruv.framework.correspondence/src/tools/vitruv/framework/correspondence/impl/InternalCorrespondenceModelImpl.xtend
```diff
@@ -253,240 +55,194 @@ class InternalCorrespondenceModelImpl extends ModelInstance implements InternalC
...
+			.filter[EcoreUtil.equals(bEObjects, resolveCorrespondingObjects(aEObjects))]
+			.flatMapFixed[removeCorrespondencesAndDependendCorrespondences].toSet
 	}
 
-	override getAllCorrespondences() {
-		return correspondences.correspondences
+	override <C extends Correspondence> Set<Correspondence> removeCorrespondencesFor(Class<C> correspondenceType,
+		Predicate<C> correspondencesFilter, List<EObject> eObjects, String tag) {
+		getCorrespondences(correspondenceType, correspondencesFilter, eObjects, tag)
+			.flatMapFixed[removeCorrespondencesAndDependendCorrespondences].toSet
```
jGleitz, line 132: This first creates a list, just to convert it into a set directly afterwards. Better:

```suggestion
			.flatMapFixedTo(new LinkedHashSet) [removeCorrespondencesAndDependendCorrespondences()]
```

bundles/framework/tools.vitruv.framework.correspondence/src/tools/vitruv/framework/correspondence/impl/InternalCorrespondenceModelImpl.xtend
```diff
@@ -253,240 +55,194 @@ class InternalCorrespondenceModelImpl extends ModelInstance implements InternalC
...
+	def private removeCorrespondence(Correspondence correspondence) {
+		EcoreUtil.remove(correspondence)
+		modified = true
+	}
+	
+	override <C extends Correspondence> Set<Correspondence> removeCorrespondencesBetween(Class<C> correspondenceType,
+		Predicate<C> correspondencesFilter, List<EObject> aEObjects, List<EObject> bEObjects, String tag) {
+		getCorrespondences(correspondenceType, correspondencesFilter, aEObjects, tag)
+			.filter[EcoreUtil.equals(bEObjects, resolveCorrespondingObjects(aEObjects))]
+			.flatMapFixed[removeCorrespondencesAndDependendCorrespondences].toSet
```
jGleitz, line 126: This first creates a list, just to convert it into a set directly afterwards. Better:

```suggestion
			.flatMapFixedTo(new LinkedHashSet) [removeCorrespondencesAndDependendCorrespondences()]
```

bundles/framework/tools.vitruv.framework.correspondence/src/tools/vitruv/framework/correspondence/impl/InternalCorrespondenceModelImpl.xtend
jGleitz: If so, I guess we can just remove this method because `getPotentiallyCachedUuid` will throw the exception for us, won’t it?

HeikoKlare: This should throw an exception and we should remove the method as soon as possible to have an exception thrown.
However, there are currently some application in which this warning occurs without a negative effect, which would fail upon the change. And even worse, the framework tests throws this warning thousands of times (see #426).
It's on my list to have the bugs in the applications and the framework tests fixed, so that this warning does not occur anymore, such that we can safely remove it.

tests/framework/tools.vitruv.framework.vsum.tests/src/tools/vitruv/framework/tests/vsum/VsumTest.java
HeikoKlare: This seems to be a leftover from when we did not have the testutils domains. I will change that.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/TestLogging.xtend
HeikoKlare: Sure, I will figure that out.

bundles/framework/tools.vitruv.framework.correspondence/src/tools/vitruv/framework/correspondence/impl/InternalCorrespondenceModelImpl.xtend
jGleitz: Ah, that’s what I feared. Then we’ll have to leave it in for now. Good that you are on it!

HeikoKlare: I've opened #434 for this issue.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/TestLogging.xtend
HeikoKlare: I've not faced an overwhelming number of message in some tests I've run. If I am not mistaken, the `UuidGeneratorAndResolverImpl` only logs warnings at one place and I would not expect that case to occur often. So I think we can safely remove that log level setting.

bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/TestLogging.xtend
```diff
@@ -43,11 +40,6 @@ class TestLogging implements BeforeAllCallback {
 		rootLogger.addAppender(new ConsoleAppender(new PatternLayout(LOG_PATTERN)))
 		rootLogger.level = ERROR
 		VITRUV_LOG_ROOTS.forEach [logger.level = toLevel(desiredLogLevel, WARN)]
-		if (System.getProperty(VM_ARGUMENT_ENABLE_ID_LOGGERS) != "true") {
-			TuidManager.logger.level = ERROR
-			TuidCalculatorAndResolverBase.logger.level = OFF
-			UuidGeneratorAndResolverImpl.logger.level = ERROR
-		}
```
jGleitz, line 50: Please remove the `VM_ARGUMENT_ENABLE_ID_LOGGERS` constant as well!

HeikoKlare: Thanks! I did not recognize that there is another constant and accidentally thought this to be tied to manually setting the log level with a VM argument. I've removed it.

HeikoKlare:
I wanted to wait until the downstream projects are prepared for this PR. But since https://github.com/vitruv-tools/Vitruv-Applications-ComponentBasedSystems/pull/131 is failing anyway and I do not know why, I will merge this PR and see whether that magically fixes that problem and otherwise investigate it with the changes in this PR.

PR #432, opened 2021-02-23
Outdated Mappings and Addresses-Recipients Demo Removal

### LINKED ISSUE #426: UUID Issue in Addresses-Recipients Demo Tests

HeikoKlare:
The PR removes the addresses-recipients demo application and outdated artifacts used for preliminary implementations of the Mappings language.
The demo application was only used for development purposes of the Mappings language and is neither complete nor technically correct. For a deeper discussion of (current) issues, see #426.

Fixes #426.


HeikoKlare:
✔️ Validated against domains/applications

PR #268, opened 2019-04-27
Support for state-based change propagation

tsaglam:
This pull request adds support for state-based change propagation: The VirtualModel is extended to support change propagation based on a new state. Every domain now holds a strategy to resolve state-based changes. A default implementation for this strategy is available, which uses EMF compare to resolve the delta-based changes. A new test project was added to test the feature.

PR #433, opened 2021-02-23
Remove EObject-based method  from StateBasedChangeResolutionStrategy interface

tsaglam:
Added default implementation for the `EObject` based method in the `StateBasedChangeResolutionStrategy` interface, so domain-specific strategies are not forced to re-implement it.


jGleitz (changes requested):
I propose to remove the method altogether.

jGleitz:
Aaaaah, GitHub swallowed my comment :angry:

bundles/framework/tools.vitruv.framework.domains/src/tools/vitruv/framework/domains/StateBasedChangeResolutionStrategy.xtend
```diff
@@ -29,5 +28,8 @@ interface StateBasedChangeResolutionStrategy {
 	 * @param uuidGeneratorAndResolver is the UUID resolver of the virtual model using this propagation strategy.
 	 * @return a {@link CompositeChange} that contains the individual change sequences.
 	 */
-	def CompositeChange<VitruviusChange> getChangeSequences(EObject newState, EObject currentState, UuidGeneratorAndResolver resolver)
+	def VitruviusChange getChangeSequences(EObject newState, EObject currentState, UuidGeneratorAndResolver resolver) {
+		return getChangeSequences(newState?.eResource, currentState?.eResource, resolver)
```
jGleitz, line 32: Okay, so here I go re-writing my long comment:

I want to question whether this method should exist in the first place. Because if I understand correctly, a StateBasedResolutionStrategy can not determine changes between two objects that are not singular roots. If I pass a non-root object or a root that is not the only root in the resource to this method, I will get too many changes, because I will receive all changes in the resource. Hence, the contract of this method is not fulfilled.

I think that if a StateBasedResolutionStrategy can only determine changes between resources, it should also only offer a method to determine changes between resources.

tsaglam: While there might be a valid discussion on whether the `EObject`-based method is required or should be removed, this interface will be adapted anyways when we abstract from resources through views. Thus, we don't need to address it now. This PR is just for convenience.

jGleitz: I don’t think the fact that code will change in the future is a valid reason to refrain from refactoring it now.

HeikoKlare: To reason about whether the method makes sense or not: What's the purpose of that method? If I see that correctly, it is only used in a test case that compare two `null` object right now.

tsaglam: I don't fully recollect why we added it, but it seems we decided to do so for some reason: https://github.com/vitruv-tools/Vitruv/pull/268#issuecomment-487592490

HeikoKlare: I do not remember either. But one of the last commits in #268 removes the EObject-based propagation from the `VirtualModel` interface, so the method within the `StateBasedChangeResolutionStrategy` maybe simply be an accidental leftover? At least, no one uses it and I think we removed the functionality from `VirtualModel` because of the reasons @jGleitz mentioned.

PR #435, opened 2021-02-23
Simplify ResourceRepositoryImpl and Avoid Reloading Models

### LINKED ISSUE #402: Remove the TransactionalChangeDomain from the VSUM
The VSUM resource set uses a `TransactionalChangeDomain`. This means that all modifications to the VSUM need to run through a command. This has a significant impact on the code base because it affects many code locations. Additionally, it leads to unexpected effects, because model modifications can technically occur always, as EMF offers no read-only model objects. For example, JaMoPP sends modification events for model objects that are still being loaded.

The only use of the `TransactionalChangeDomain` seems to be in `ChangePropagatorImpl`, where the changed objects are determined through the domain. This code should be changed to get the information from the changes that were recorded in the VSUM.

After this modification, all code related to `TransactionalChangeDomain` and `VitruviusRecordingCommand` can be removed, which should be a significant simplification. It would also close #382.

HeikoKlare:
This PR simplified the `ResourceRepositoryImpl`. It removes all logic related to accidentally loading and reloading models due to incomplete change sequences. It leads to a `ModelInstance` that degrades to a wrapper for resources, which can/will also be removed.

The new logic is as simple as follows: A model will be _loaded_ if it the V-SUM is reloaded (currently not fully implemented) or if it is a `pathmap` resource. All other models will be retrieved from the resource set, if they are already in there, or otherwise created.

This PR also fixes bugs regarding UUID resolution: Pathmap resources were loaded by normalizing their URIs before, which (somehow) led to problems with their UUID resolution (apart from the unnecessity to not have them loaded as actual pathmap resources). If I remember correctly, we made this because otherwise Java resources which shall actually be modified were loaded as pathmap resources (such as a compilation unit referenced from a package is loaded via a pathmap resource during proxy resolution). However, since we now properly create models from scratch within the V-SUM and do not perform any reload magic, this does not seem to be a problem anymore and we can safely remove it.

Fixes #402.


bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.xtend
```diff
@@ -51,160 +43,101 @@ class ResourceRepositoryImpl implements ModelRepository {
...
-			// model is not existing yet
-			logger.info('''Exception during loading of model instance «modelInstance» occured: «re»''')
+	override getModel(VURI modelURI) {
+		val existingInstance = modelInstances.get(modelURI)
+		if (existingInstance !== null) {
+			return existingInstance
 		}
-
+		modelURI.createOrLoadModel(false)
+	}
```
jGleitz, line 56: A `ModelInstance` holds no state of its own. I feel like it’s wasteful to build a cache `VURI` -> `ModelInstance` when this just duplicates the cache that is already in the `ResourceSet`. Can’t we simply hand out new instances of `ModelInstance` for each invocation of `getModel`?

HeikoKlare: It's not only a cache but also needed to determine which models have to be saved, which have to be deleted and finally whose URIs have to be stored to persist a V-SUM. We could also derive that information from the resource set, but then we need to wrap the resources into model instances again for saving or deleting them, which would actually make the class completely obsolete.
In addition, model instances are retrieved quite often, so I expect the "cache" to outweigh the overhead of element creation.

jGleitz: > We could also derive that information from the resource set

Yes, that is what I am proposing.

> which would actually make the class completely obsolete.

Which class would be obsolete? `ResourceRepositoryImpl` would still be a meaningful abstraction over a resource set, and `ModelInstance` would still be a meaningful abstraction over a resource, wouldn’t they?

> In addition, model instances are retrieved quite often, so I expect the "cache" to outweigh the overhead of element creation.

I doubt that, element creation is _very_ cheap and the general advice is that object pooling has no performance benefit on the JVM as long as there is no other resource tied to the objects. (see, for example, https://softwareengineering.stackexchange.com/a/115166/185658)

HeikoKlare: I have to admit that except for the fact that the resource set contains more resources than the models (in particular, the uuid resource), we could do as you propose (although I do not see as strong benefits for one solution as you seem to see).

My biggest point against this, however, is the way how resource sets behave. I have been working with them for years and it's still like seeing them the first time when trying to do something with them. I quickly wanted to implement your proposal to see how the code looks like, but I don't get it to work. Why is it the case? I expected `getResource(uri, false)` to return nothing if the resource at `uri` is not yet loaded into the resource set. This is, however, _not_ the case. Pathmap resources  may still be loaded, because someone else has loaded them and then some global registry contains them from which they are retrieved. So the assumption that "only perform initialization when not in the set" works is wrong. I can, of course, adapt the code to work around this strange behavior, but my fear is that it results in some strange code that no one deeply understanding resource sets comprehends well (including me) and which is harder to maintain than the simple (and well behaving) cache that we have now.
This is why I am still in favor of keeping the additional structure (although I see your points against ist).

jGleitz: My main point against the additional structure was that it is more complicated and more error-prone to maintain it. If, however, using the resource set is actually more complicated and error-prone, my main argument becomes void. So I think it’s reasonable to leave things as they are for now.

HeikoKlare: > My main point against the additional structure was that it is more complicated and more error-prone to maintain it.

I absolutely agree to this and found the solution I implemented without the additional structure to better in the first place. But in the current implementaiton I find the functionality for the different cases (new resource creation vs. loading existing resource) easy to comprehend, while it get pretty bloaded because of exceptional behavior I have mentioned.

I will check whether this PR works well with our applications and if it does so, I would merge it if you agree?
I hope that these simplifications make it even easier to integrate your modifications for supporting changes to multiple resources.

jGleitz: > I would merge it if you agree

Go ahead!

PR #436, opened 2021-02-24
UuidGeneratorAndResolver Non-Strict Mode Removal

### LINKED ISSUE #195: UUID Strict Mode in EMF Editor Monitor
Strict mode for UUID resolution should only be disabled in tests to be able to monitor models from an arbitrary state instead of from its initial creation. We currently have to disable the strict mode in `EMFModelChangeRecordingEditorSaveListener` because otherwise transient elements are not handled correctly (e.g. the `DocumentRoot` element in CAEX models). Deactivation of the strict mode for editor monitors was reintroduced with PR #194.

We have to investivate why the mechanism does not work properly with transient elements (or whatever the reason for failing with CAEX models is) and fix it so that strict mode can be properly enabled again.

HeikoKlare:
Removes the non-strict mode of the `UuidGeneratorAndResolverImpl`. It was able to tolerate non-existing UUIDs by setting the `strictMode` flag to `false` in the constructor. This was especially necessary for the `VirtualModel`, because _incomplete_ domains, i.e., domains in which no complete change sequences are tracked, reloaded resources in which elements occurred that did not have a UUID. Since we have removed that workaround, the resolver can run in strict mode also within the `VirtualModel`. Since no one else needs the non-strict mode, this PR completely removes it.

Apparently, the non-strict mode was used in some cases in monitors (see #195). We should, however, adapt the monitor there and not keep the non-strict mode that can hide further bugs.

Fixes #195.


HeikoKlare:
✔️ works fine with our applications

PR #438, opened 2021-02-24
Make UUIDs Mandatory

### LINKED ISSUE #434: UuidGeneratorAndResolver Swallows Missing UUIDs
The `UuidGeneratorAndResolverImpl` currently swallows exceptions when UUIDs are missing. Since the cases in our case studies in which the UUIDs are missing are not problematic, this does not yet lead to follow-up errors. However, this behavior is not as intended, because every object has to have a UUID. Thus, we need to investigate the cases in which exceptions are swallowed (printing only a warning) and fix them, such that we can afterwards remove these lines from the `UuidGeneratorAndResolverImpl`.

https://github.com/vitruv-tools/Vitruv/blob/249f82bd0d84d6b888411ffb43a6734916751a6f/bundles/framework/tools.vitruv.framework.correspondence/src/tools/vitruv/framework/correspondence/impl/InternalCorrespondenceModelImpl.xtend#L151-L158

HeikoKlare:
This PR makes UUIDs mandatory. It adapts the `UuidGeneratorAndResolveImpl` such that it throws an exception when trying to register an `EObject` that does not have a UUID yet. In addition, it adapts the `InternalCorrespondenceModelImpl` such that it expects elements to have a UUID an otherwise throws an exception.

Fixes #434.


bundles/framework/tools.vitruv.framework.correspondence/src/tools/vitruv/framework/correspondence/impl/InternalCorrespondenceModelImpl.xtend
```diff
@@ -134,22 +134,11 @@ class InternalCorrespondenceModelImpl extends ModelInstance implements InternalC
...
-	}
-
 	override <C extends Correspondence> Set<C> getCorrespondences(Class<C> correspondenceType,
 		Predicate<C> correspondencesFilter, List<EObject> eObjects, String tag) {
 		val correspondences = new HashSet<Correspondence>()
-		if (eObjects.haveUuids) {
-			val uuids = eObjects.map[uuidResolver.getPotentiallyCachedUuid(it)]
-			correspondences += getCorrespondencesForUuids(uuids)
-		}
+		val uuids = eObjects.map[uuidResolver.getPotentiallyCachedUuid(it)]
```
jGleitz, line 140:
```suggestion
		val uuids = eObjects.mapFixed [uuidResolver.getPotentiallyCachedUuid(it)]
```

The list will be iterated multiple times by `getCorrespondencesForUuids` but the result of `map` is not cached.

jGleitz (changes requested):

HeikoKlare:
The `ChangeRecorderTest`s are failing, because they use the inaccuracy that the `UuidGeneratorAndResolverImpl` ignores that an element may not have a UUID when trying to register it.
Currently, it is necessary to register an object at the `ChangeRecorder` (with `addToRecording`) to ensure that for its added child elements UUIDs are generated. Calling `addToRecording` for an existing element that has no UUID assigned leads to the situation that a UUID for that element is missing. This is not as intended. We can do the following:
1. Enforce the implicit contract and update the `ChangeRecorderTest`. Then, however, monitoring will not be possible for `EObject`s, unless the element was already monitored and is reloaded into a child resolver, because the `EObject` to be added to monitoring does not have a UUID yet. We could the also change `addToRecording` to only accept resources instead of `EObject`s.
2. Allow recording of `EObject`s by creating a UUID for the element added to recording if it does not exist yet and only enforce the existence of UUIDs for the child elements. Then you can add a root element to the recorder upon creation and UUIDs for its childs will be generated on their insertion.
3. Completely remove the assumption that UUIDs exist for elements added to recording (like is done now by the incomplete implementation in the `UuidGeneratorAndresolverImpl`. Then, however, we cannot distinguish whether the absence of UUIDs is intended or not.

I am in favor of 1 or 2, because 3 can easily hide bugs. In addition, all our applications work fine with the change proposed in this PR, but the tests cover scenarios that are currently not required (monitoring `EObject`s  rather than resources).

jGleitz:
I think 1. sounds reasonable. The contract for 2. is difficult to understand.

1. is an extension of Vitruv’s assumption that it will see the creation of any object it ever sees. While this assumption is not a good one, I think it is consequential to also use it in `ChangeRecorder` as long as Vitruv makes it anyway.

HeikoKlare:
One further drawback of 1 is that it means that the `ModelRepositoryImpl` will never work as it is then impossible to monitor elements rather than resources. But probably that is fine and we should completely throw away the idea of having a `ModelRepository`.

jGleitz:
> Why should that assumption be bad?

That’s a personal opinion. It makes Vitruv a monolithic tool that has _always_ to be involved per assumption. That, in turn, makes Vitruv much less useful because it cannot be used in an ad-hoc fashion. My wager is that should somebody ever try to use Vitruv in practice, the assumption that Vitruv has seen every creation will make a lot of problems.

> it means that the ModelRepositoryImpl will never work as it is then impossible to monitor elements

 1. The idea of abstracting away resources is attractive, but also feels like an uphill battle because resources are so central to EMF.
 2. I don’t think that the decision would make it impossible. We would just need to add a method like `addNewlyCreatedToRecording` that first creates a create change for the object and then starts recording on it. This would cover the use case of monitoring a new element and would create the UUIDs.

HeikoKlare:
>That, in turn, makes Vitruv much less useful because it cannot be used in an ad-hoc fashion. 

Alright, so that's not a drawback of our implementation but of the concept in general. Of course, this is the essential drawback of an inductive approach, like we discuss in our papers, which is why we have worked on integration mechanisms as well. Throwing the assumption away has other drawbacks, but this shouldn't end in a fundamental discussion about research objectives.

> The idea of abstracting away resources is attractive, but also feels like an uphill battle because resources are so central to EMF.

That's true. I think the root cause of our way towards throwing away resources was the sometimes cumbersome handling within JaMoPP. In particular, one problem within the `ResourceRepositoryImpl` was that compilation units occur twice, once within the package resource and once within their own resource (using containment proxies). This led to duplicate monitoring, which is why we tried to work with elements rather than resources. But this is obviously something that we can solve in a different or, as far as I see, do not need to solve anymore, but our change recorder was already updated accordingly.

This PR will become larger, because the discussed changes also reveal some shortcoming of the change recorder that I think we have already discussed before: the `NotificationToEChangeRecorder` does, by accident, not assume elements to have UUIDs assigned. I think this was necessary because of all the workarounds for incompletely tracked domains in the V-SUM, but this is something we can and should throw away now to avoid that it hides bugs. Again, changing this does not seem to produce any problems within the applications but only without our unit tests, as they do not make the assumptions which the applications correctly make.

HeikoKlare:
Depends on https://github.com/vitruv-tools/Vitruv/pull/442 to keep our applications working, as some tests are only working properly by accident (UUIDs are accidentally removed, but since they are not mandatory this does not lead to an error).

jGleitz:
This PR removes a lot of separating spaces from `ChangeRecorder` and I think that makes its formatting worse (you probably applied the Formatter)

HeikoKlare:
I have to admit that I don't see any changes that make formatting worse. Which kinds of separating spaces do you mean, the ones for lambda expressions? This is ordinary Xtend formatting. As you mentioned, I guess that I have applied the formatter.

jGleitz:
I find those separating spaces improve clarity. But yeah, I guess it’s a matter of taste :shrug:.

HeikoKlare:
For me, removing them is fine. Why do you think they improve clarity? At least you cannot mix them up with array indexing, as it does not exist in Xtend.

Anyway, this PR works with the applications and I'll merge it now to finally enforce proper UUID management in the `VirtualModel`, so that we can proceed with further issues.

PR #439, opened 2021-02-24
Remove ModelRepositoryImpl and Cleanup Responsibilities in V-SUM

### LINKED ISSUE #58: Manage models instead of resources in VSUM
The VSUM or more specifically the ModelRepository manages so called ModelInstances. These ModelInstances internally rely on EMF resources.
This implementation has two drawbacks:
1. Resources creation is delayed and not performed just-in-time. So if a Reaction creates a new model, the VSUM will first know about that model when it gets persisted after processing all performed changes. In fact, the VSUM should know about a new model just in the moment it gets created, especially because it will have to monitor newly created models for changes in the future
2. The VSUM manages physical instead of logical models. The Java model, for example, consists of the root package and all other packages and compilation units are contained within that root packages. Therefore, the VSUM should know about that root package that contains all other model information. Physically, the model is persisted in several resources (one for each package and compilation unit), but that is a secondary information attached to the model elements.

We should change the implementation, so that the VSUM primarily knows about logical models (elements that have not container) and manages the resources as secondary artifacts that are potentially still created/modified delayed.

HeikoKlare:
This PR removes the `ModelRepositoryImpl` according to the discussion in #438. It streamlines the responsibilities within the `vsum` project, such that the `ResourceRepositoryImpl` is responsible for all managing and saving all kinds of resources. This also reduces the necessary number of instances passed to the `ChangePropagatorImpl`. Finally, it also reduces several `UuidGeneratorAndResolver` dependencies to `UuidResolver` dependencies.

Closes #58.


bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.xtend
```diff
@@ -26,15 +26,21 @@ import tools.vitruv.framework.vsum.helper.VsumFileSystemLayout
...
 	static val logger = Logger.getLogger(ResourceRepositoryImpl)
 	val ResourceSet resourceSet
 	val VitruvDomainRepository domainRepository
 	val Map<VURI, ModelInstance> modelInstances = new HashMap()
 	val VsumFileSystemLayout fileSystemLayout
-	@Accessors
 	val UuidGeneratorAndResolver uuidGeneratorAndResolver
+	override getUuidResolver() {
+		return uuidGeneratorAndResolver
+	}
```
jGleitz, line 41: This is a very odd placement for this method, I’d prefer if it was pushed further down

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.xtend
```diff
@@ -126,6 +133,16 @@ class ResourceRepositoryImpl implements ModelRepository {
 		var Resource uuidProviderResource = resourceSet.loadOrCreateResource(uuidProviderVURI.EMFUri)
 		new UuidGeneratorAndResolverImpl(this.resourceSet, uuidProviderResource)
 	}
+	
+	def private initializeCorrespondenceModel() {
+		var correspondencesVURI = fileSystemLayout.correspondencesVURI
+		logger.trace('''Creating or loading correspondence model from: «correspondencesVURI»''')
+		val correspondencesResource = resourceSet.withGlobalFactories().loadOrCreateResource(
```
jGleitz, line 140: This puts the correspondences back into the main resource set. We wanted to avoid that because Correspondences don’t need UUIDs and changes should also not be recorded for them.

jGleitz (commented):
I am not convinced of moving the correspondence model to the resource repository. Because: 

 1. My intuitive understanding of the resource repository is that it manages the data resources, not the metadata resources. So while the move makes it easier to reference the correspondence model, it feels like the worse separation of concerns.
 2. To strengthen the previous point: As I’ve already said, there can conceptually be a resource repository without a correspondence model, so that alone makes it feel weird that a resource repository has a dependency on the correspondence model.
 2. The correspondences should live in a different resource set, from my point of view.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.xtend
HeikoKlare: Yes, of course.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.xtend
HeikoKlare: They should neither get UUIDs assigned nor get recorded. They are simply another _resource_ to be managed and stored.

jGleitz: I have not checked with a debugger, but since the UUID resolver is registered before loading the correspondences, we should run through the `loadUuidsIntoChild` code, shouldn’t we?

Regardless of whether UUIDs actually get assigned, I still think that there is no advantage of having the UUID resource in the data resource set (there are no references to other objects). However, I see the disadvantage that it makes the code more prone to bugs: we have more non-data resources in the resource set, and we could accidentally start recording UUIDs for correspondences again in the future. We probably wouldn’t even notice, however, correspondences can make up to 50% of the uuid repository in practice.

HeikoKlare: Regarding the former point: First, we can load the correspondences before registering the UUID resolver. However, `loadUuidsIntoChild` will do nothing, because the `parentResolver` is empty anyway. There is no parent to load UUIDs from. Second, even without changing the order, there is no recorder registered for the correspondence resource, so no UUIDs will be generated later on. At least, I don't see why this should be the case.

> having the UUID resource in the data resource set (there are no references to other objects)

That's not true. The UUID resource references all the objects of the models in the resource set.

> we have more non-data resources in the resource set,

I have to admit that I do not know what _non-data resources_ are.

> we could accidentally start recording UUIDs for correspondences again in the future

We can easily provide a test that ensures that we don't do that. I think the problem is that you are thinking of the `ResourceRepositoryImpl` as an `OnlyContainsModelsAccessedByUsersRepository` or `ResourcesRequiringUuidsRepository` while I think of it as a `ResourcesOfaVsumRepository`. I don't think that whether resources may reference each other or not is a valid metrics for identifying the borders of resource sets, because then we should have one resource set per domain and simply start a single recorder for each of these resource sets. But I have to admit that I don't see the benefit of having that (like I don't see the benefit of separating the correspondence model).

jGleitz: > having the UUID resource in the data resource set (there are no references to other objects)

That was meant to read ‘having the _correspondence_ resource in the data resource set

> I do not know what non-data resources 

‘Metadata resources’ would be a better term.

> I don't think that whether resources may reference each other or not is a valid metrics for identifying the borders of resource sets

It is not the _only_ argument for whether resources should be in the same resource set, but I certainly think it is a _valid_ argument. On one hand, all resources that _do_ reference each other must be in the same resource set. On the other hand, being in the same resource set can have side effects, because of adapters that might be installed in the resource set. That gives me a bias of wanting to separate resources into different resource sets whenever feasible.

> because then we should have one resource set per domain and simply start a single recorder for each of these resource sets.

Exactly! I think if we could do this it would actually be a more elegant way to manage entities in Vitruv. The only reason I haven’t proposed a PR doing exactly that yet is that the UUID persistence would need to change.

---

> you are thinking of […] `OnlyContainsModelsAccessedByUsersRepository` or `ResourcesRequiringUuidsRepository` while I think of it as a `ResourcesOfaVsumRepository`

Yep, that’s exactly the difference. And I’d propose that my view has a clearer separation of concerns, while your view mixes concerns without any advantage that I could see.

---

Long story short: There compulsory argument against the architecture in this PR. However, I think that the architecture could be improved by separating concerns more clearly.

HeikoKlare: > ‘Metadata resources’ would be a better term.

Right now, correspondences may seem to be metadata, but actually they are not. I contrast to, for example, UUIDs, correspondences will be used to derive views, because they are (or can be) necessary elements to witness relations between elements that have to be considered by or can be represented in views. However, whether or not we consider them metadata may not make that much a difference.

> The only reason I haven’t proposed a PR doing exactly that yet is that the UUID persistence would need to change.

As soon as we have "proper views", it should be easy to establish one `UuidGeneratorAndResolver` per resource set (i.e., per domain). However, you are right that it's not that easy right now.

> Yep, that’s exactly the difference. And I’d propose that my view has a clearer separation of concerns, while your view mixes concerns without any advantage that I could see.

More precisely, I've seen the advantage in having the correspondence model in the `ResourceRepositoryImpl` but not necessarily within the single resource set managed by it. So what about having the correspondence model managed in the `ResourceRepositoryImpl` but not have it in the same resource set as the other models? Then we have all (kinds of) models wrapped in that repository, saved at the same time and so on, but not mixed concerns within one resource set. As soon as we separate resource sets by domains (like discussed above), there will be multiple resource sets to be managed anyway and we probably wont establish one `ResourceRepository` for each of them.

jGleitz: >  correspondences may seem to be metadata, but actually they are not

We seem to have different notions of ‘metadata’. I used the term simply to mean ‘data about data’.

> So what about having the correspondence model managed in the ResourceRepositoryImpl but not have it in the same resource set as the other models

Seems like a sensible common ground. It addresses my objection that having the correspondences in the same resource set might lead to hidden bugs. And it has all resources managed by the same class, which you see as being advantageous.

HeikoKlare: > I used the term simply to mean ‘data about data’.

Me too, but correspondences are data and not data about data ;-) 

> Seems like a sensible common ground. 

Perfect, I'll propose an according update.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.xtend
HeikoKlare: Done.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/repositories/ResourceRepositoryImpl.xtend
HeikoKlare: I've separated the correspondence model into a dedicated resource set and documented the ideas for enhancement in #440.

HeikoKlare:

PR #441, opened 2021-03-01
Make EObject References in UUID Repository Derived

HeikoKlare:
In the `UuidGeneratoraAndResolverImpl`, an Ecore-based model is used for mapping UUIDs to `EObjects` and vice versa. Since this is handled as an ordinary Ecore model, removing elements from its references leads to the removal from this UUID repository. This will always occur when using the cross-referencer to remove elements from references, such as calling `destroy()` on UML model elements. To avoid this, the PR sets the `EObject` references in this repository to as `derived`, which avoids that their values are replaced from anywhere else than from our code. Since we define the entries of the repository map anyway, this change does not degrade operation of the repository.


HeikoKlare:

HeikoKlare:
Will be replaced by #442.

Issue #440, opened 2021-02-24
One Resource Set per Domain in ResourceRepositoryImpl


HeikoKlare:
The `ResourceRepositoryImpl` currently manages a resource set containing all domain models. Only the correspondence model is separated into an additional resource set. However, for each resource a proper domain recorder has to be registered, because different recorders track changes for the different domains. It would be beneficial to manage one resource set per domain such that we can register a recorder for the complete resource set instead of deciding for each resource on its own. This is currently problematic because the `UuidGeneratorAndResolverImpl` uses a resource containing the UUIDs and references to the according `EObjects` and thus has to be placed in the domain models' resource set. Having one resource set per domain would best come along with one `UuidGeneratorAndResolverImpl` per domain. This will be possible as soon as UUID handling is hidden in the V-SUM and no resolver is provided by the `VirtualModel` anymore. Then, it should be possible to perform the discussed separation of domain models into domain-specific resource set.
See also the discussion in #439.

JanWittler:
I would propose to close this issue as it is outdated? As we removed the domains, I don't think there is anything left to do here @HeikoKlare ?
With recent refactoring, the `ResourceRepositoryImpl` now has different recorders based on file ending instead of domains.

HeikoKlare:
Actually, the domains were only used to retrieved file extensions that belong together to instantiate a common change recorder. Thus, we could still address this issue by setting up one resource set per file extension. The question is, however, whether that enhancement is worth the effort, since we currently have no problems putting all models into the same resource set. It would just simplify the functionality for attaching change recorders to resources, as we can then simply monitor a complete resource set rather than monitoring resource additions to apply the appropriate recorder.

JanWittler:
Having a look at the source code, I am wondering why we even use different recorders for each file extension? Why not one recorder for the entire resource set? Even if the assumption holds that each file extension uniquely identifies a domain (which does obviously not), what would be the benefit of having one change recorder per domain instead of one for all domains?

Anyways, I would strongly vote against having one resource set per file extension, as this will lead to problems with element resolution as soon as a model references another element in a model with a different file extension.

HeikoKlare:
True, so I absolutely agree to closing this issue.

[...]

PR #442, opened 2021-03-01
UUID Resolver Element Cleanup and Placement in Separate ResourceSet

### LINKED ISSUE #298: UUID model persistence
It seems like the UUID model is currently not persisted (in the `uuid` folder of the V-SUM project), although a resource is properly created for it within the `UuidGeneratorAndResolverImpl`.
In the `ResourceRepositoryImpl`, there is only explicit persistence for the correspondence model but not for the UUID model, which may be the problem.

HeikoKlare:
This PR adds functionality for cleaning up and saving the `UuidGeneratorAndResolverImpl` while separating it from models resources and effects to them. In detail it provides the following:
* Functionality for cleaning up removed elements in `UuidGeneratorAndResolverImpl`, i.e., removing elements that are not placed in a resource anymore, which is automatically executed on saving the instance.
* Placement of the Uuid resource in a separate resource set, which was placed in the models' resource set before. This is useful because some Ecore functionality (such as `EcoreUtil.delete`) uses the cross referencer for a complete resource set, thus unintentionally affecting the UUID model. For example, calling `EcoreUtil.delete` on a model object will also remove it from the repository of the UUID resolver if placed in the same resource set. See also #441, for which this is a replacement. In addition, the resource in now managed within the `UuidGeneratorAndResolverImpl` (as it uses a separate resource set anyway).
* Have the resource of the `UuidGeneratorAndResolver` automatically saved (and cleaned up on save) in the `ResourceRepositoryImpl`. This only possible because of element cleanup, as otherwise orphan elements without a resource prevent the serialization of the resource.
* Proper implementation of load functionality for the `UuidGeneratorAndResolverImpl`, resolving the elements in the models' resource set. This also serves as a loading functionality for the model resources of a `VirtualModel`, such that storing them separately is not necessary. The UUID storage contains all necessary information about model resources to be loaded.
* Addition of UUID test cases for element movement, element deletion, resolver cleanup and saving, as well as different loading scenarios.

Replaces and closes #441.
Fixes #298.

### DIFF
--- a/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
+++ b/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
@@ -10,7 +10,7 @@ import org.eclipse.emf.ecore.EObject
  * It registers a notifier on that {@link ResourceSet}, which informs about newly added/loaded {@link Resource}s
  * and automatically loads the assigned {@link Uuid}s for contained elements from the parent resolver.
  */
-interface UuidGeneratorAndResolver extends UuidResolver {
+interface UuidGeneratorAndResolver extends UuidResolver, AutoCloseable {
 	/**
 	 * Registers an object and returns the generated UUID for it.
 	 */
--- a/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
+++ b/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
@@ -15,11 +15,16 @@ interface UuidGeneratorAndResolver extends UuidResolver {
 	 * Registers an object and returns the generated UUID for it.
 	 */
 	def String generateUuid(EObject eObject);
-	
+
 	/**
 	 * Registers an object that was not created before and thus has no UUID.
 	 * This is only successful if the element is globally accessible (third party element) or if
 	 * we are not in strict mode. Otherwise an exception is thrown, because a previous create is missing. 
 	 */
 	def String generateUuidWithoutCreate(EObject eObject);
+
+	/**
+	 * Cleans up elements that have been removed, i.e., that are not contained in a resource anymore
+	 */
+	def void cleanupRemovedElements()
 }
--- a/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
+++ b/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
@@ -24,7 +24,7 @@ interface UuidGeneratorAndResolver extends UuidResolver {
 	def String generateUuidWithoutCreate(EObject eObject);
 
 	/**
-	 * Cleans up elements that have been removed, i.e., that are not contained in a resource anymore
+	 * Saves the mapping between {@link Uuids}s and {@link EObject}s to an underlying resource.
 	 */
-	def void cleanupRemovedElements()
+	def void save()
 }
--- a/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
+++ b/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
@@ -62,7 +62,8 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 	 * @param resourceSet -
 	 * 		the {@link ResourceSet} to load model elements from, may not be null
 	 * @param resourceUri -
-	 * 		the {@link URI} to place a resource for storing the mapping in, may be null
+	 * 		the {@link URI} to persist this repository to when {@link #save} is called. If {@code null}, this 
+	 * 		UUID repository will not be persisted.
 	 * @throws IllegalArgumentException if given {@link ResourceSet} is null
 	 */
 	new(ResourceSet resourceSet, URI resourceUri) {
@@ -78,7 +79,8 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 	 * @param resourceSet -
 	 * 		the {@link ResourceSet} to load model elements from, may not be null
 	 * @param resourceUri -
-	 * 		the {@link URI} to place a resource for storing the mapping in, may be null
+	 * 		the {@link URI} to persist this repository to when {@link #save} is called. If {@code null}, this 
+	 * 		UUID repository will not be persisted.
 	 * @throws IllegalArgumentException if given {@link ResourceSet} is null
 	 */
 	new(UuidResolver parentUuidResolver, ResourceSet resourceSet, URI resourceUri) {
--- a/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
+++ b/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
@@ -15,16 +15,19 @@ import static extension edu.kit.ipd.sdq.commons.util.org.eclipse.emf.common.util
 import static com.google.common.base.Preconditions.checkArgument
 import static com.google.common.base.Preconditions.checkState
 import tools.vitruv.framework.util.ResourceRegistrationAdapter
+import org.eclipse.emf.ecore.resource.impl.ResourceSetImpl
+import static extension tools.vitruv.framework.util.ResourceSetUtil.withGlobalFactories
 
 /**
  * {@link UuidGeneratorAndResolver}
  */
 class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 	static val logger = Logger.getLogger(UuidGeneratorAndResolverImpl)
-	final ResourceSet resourceSet
-	final UuidResolver parentUuidResolver
-	UuidToEObjectRepository repository
-	UuidToEObjectRepository cache
+	val ResourceSet resourceSet
+	val Resource uuidResource
+	val UuidResolver parentUuidResolver
+	val UuidToEObjectRepository repository
+	val UuidToEObjectRepository cache
 
 	/**
 	 * Instantiates a UUID generator and resolver with no parent resolver, 
@@ -55,44 +58,41 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 	/**
 	 * Instantiates a UUID generator and resolver with no parent resolver, 
 	 * the given {@link ResourceSet} for resolving objects
-	 * and the given {@link Resource} for storing the mapping.
+	 * and a resource at the given {@link URI} for storing the mapping in.
 	 * @param resourceSet -
 	 * 		the {@link ResourceSet} to load model elements from, may not be null
-	 * @param uuidResource -
-	 * 		the {@link Resource} to store the mapping in, may be null
+	 * @param resourceUri -
+	 * 		the {@link URI} to place a resource for storing the mapping in, may be null
 	 * @throws IllegalArgumentException if given {@link ResourceSet} is null
 	 */
-	new(ResourceSet resourceSet, Resource uuidResource) {
-		this(null, resourceSet, uuidResource)
+	new(ResourceSet resourceSet, URI resourceUri) {
+		this(null, resourceSet, resourceUri)
 	}
 
 	/**
 	 * Instantiates a UUID generator and resolver with the given parent resolver, used when
 	 * this resolver cannot resolve a UUID, the given {@link ResourceSet} for resolving objects
-	 * and the given {@link Resource} for storing the mapping.
+	 * and a resource at the given {@link URI} for storing the mapping in.
 	 * @param parentUuidResolver -
 	 * 		the parent {@link UuidResolver} used to resolve UUID if this contains no appropriate mapping, may be null
 	 * @param resourceSet -
 	 * 		the {@link ResourceSet} to load model elements from, may not be null
-	 * @param uuidResource -
-	 * 		the {@link Resource} to store the mapping in, may be null
+	 * @param resourceUri -
+	 * 		the {@link URI} to place a resource for storing the mapping in, may be null
 	 * @throws IllegalArgumentException if given {@link ResourceSet} is null
 	 */
-	new(UuidResolver parentUuidResolver, ResourceSet resourceSet, Resource uuidResource) {
+	new(UuidResolver parentUuidResolver, ResourceSet resourceSet, URI resourceUri) {
 		checkArgument(resourceSet !== null, "Resource set may not be null")
 		this.resourceSet = resourceSet
-		if (uuidResource?.resourceSet == resourceSet) {
-			// Using the same resource set for models and the UUID resource can lead to accidental
-			// UUID removals, such as EcoreUtil.delete(element) removing the UUID mapping for element
-			logger.warn("UUID resource should not be placed into the same resource set as the models")
-		}
 		this.parentUuidResolver = parentUuidResolver ?: UuidResolver.EMPTY
-		loadAndRegisterUuidProviderAndResolver(uuidResource)
+		this.uuidResource = if (resourceUri !== null) new ResourceSetImpl().withGlobalFactories.loadOrCreateResource(resourceUri)
+		this.repository = loadOrGenerateUuidRepository(uuidResource)
+		this.cache = UuidFactory.eINSTANCE.createUuidToEObjectRepository
 		this.resourceSet.eAdapters += new ResourceRegistrationAdapter[resource|loadUuidsFromParent(resource)]
 	}
-
-	def private loadAndRegisterUuidProviderAndResolver(Resource uuidResource) {
-		var UuidToEObjectRepository repository = uuidResource?.resourceContentRootIfUnique
+	
+	def private loadOrGenerateUuidRepository(Resource uuidResource) {
+		var repository = uuidResource?.resourceContentRootIfUnique
 			?.dynamicCast(UuidToEObjectRepository, "uuid provider and resolver model")
 		if (repository === null) {
 			repository = UuidFactory.eINSTANCE.createUuidToEObjectRepository
@@ -100,8 +100,7 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 				uuidResource.contents += repository
 			}
 		}
-		this.repository = repository
-		this.cache = UuidFactory.eINSTANCE.createUuidToEObjectRepository
+		return repository
 	}
 
 	override getUuid(EObject eObject) {
@@ -315,8 +314,13 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 		}
 	}
 	
-	override cleanupRemovedElements() {
-		for (val iterator = repository.EObjectToUuid.keySet.iterator(); iterator.hasNext(); ) {
+	override save() {
+		cleanupRemovedElements()
+		uuidResource?.save(null)
+	}
+	
+	private def cleanupRemovedElements() {
+		for (val iterator = repository.EObjectToUuid.keySet.iterator(); iterator.hasNext();) {
 			val object = iterator.next()
 			if (object.eResource === null) {
 				val uuid = repository.EObjectToUuid.get(object)
--- a/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
+++ b/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
@@ -310,4 +310,15 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 		}
 	}
 	
+	override cleanupRemovedElements() {
+		for (val iterator = repository.EObjectToUuid.keySet.iterator(); iterator.hasNext(); ) {
+			val object = iterator.next()
+			if (object.eResource === null) {
+				val uuid = repository.EObjectToUuid.get(object)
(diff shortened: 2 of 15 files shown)


bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -81,6 +81,11 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 	new(UuidResolver parentUuidResolver, ResourceSet resourceSet, Resource uuidResource) {
 		checkArgument(resourceSet !== null, "Resource set may not be null")
 		this.resourceSet = resourceSet
+		if (uuidResource?.resourceSet == resourceSet) {
+			// Using the same resource set for models and the UUID resource can lead to accidental
+			// UUID removals, such as EcoreUtil.delete(element) removing the UUID mapping for element
+			logger.warn("UUID resource should not be placed into the same resource set as the models")
+		}
```
jGleitz, line 88: Is there a good reason not to make this an exception? If there is no valid reason to do this, I think we should forbid it. Only logging a warning opens the possibility for one developer thinking ‘it doesn’t matter in this case’ and the next developer wondering where some odd behaviour comes from.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
```diff
@@ -15,11 +15,16 @@ interface UuidGeneratorAndResolver extends UuidResolver {
...
 	 * Registers an object that was not created before and thus has no UUID.
 	 * This is only successful if the element is globally accessible (third party element) or if
 	 * we are not in strict mode. Otherwise an exception is thrown, because a previous create is missing. 
 	 */
 	def String generateUuidWithoutCreate(EObject eObject);
+
+	/**
+	 * Cleans up elements that have been removed, i.e., that are not contained in a resource anymore
+	 */
+	def void cleanupRemovedElements()
```
jGleitz, line 29: The method exposes internals on the API. I don’t think that `cleanupRemovedElements` is something a client would ever have a reason to call. True, `UuidGeneratorAndResolver` already exposes several internals, but that’s not a good reason to make it worse.

To improve this, the easiest option would be to introduce something like a `beforeSave` hook, which has a much simpler contract (i.e. call me before saving me). But then again, I think it’s not good that some external entity serializes the internals of `UuidGeneratorAndResolver`.

So probably the cleanest way would be to move the `save` functionality to `UuidGeneratorAndResolverImpl` altogether. This way, we have true information hiding for the internal state of it.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: There is a reason, but it's a valid question whether it is a _good_ one. My intention was that the only reason not to place the UUID resource in the models' resource set is that resource-set-spanning actions can unintentionally (and erroneously) affect the UUIDs. If you are able to avoid that situation, you may also put the UUID resource in the same resource set (so it's not always wrong to place it there). But since there is no benefit in having the UUID resource in the same resource set as the models and, as you point out, there is the risk of misusing the possibility to do so, I agree that we better should have this throw an exception instead.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
HeikoKlare: > `UuidGeneratorAndResolver` already exposes several internals, but that’s not a good reason to make it worse.

It's true that we should spend some time into refactoring the `UuidGeneratorAndResolver` and its interface. I also expect that we can improve information hiding, especially because the reduced number of workarounds probably makes some of the logic even obsolete, which we had to expose to make them work before.

I do not want to argue whether `cleanupRemovedElements` should be exposed or not. I think in combination with putting the `save` functionality to the `UuidGeneratorAndResolverImpl`, it makes sense anyway (assuming that cleanup is always tied to saving). By now, the resource handling was placed outside the `UuidGeneratorAndResolverImpl`, which becomes obsolete when placing the resource in a separate resource set anyway. I would thus propose to completely move the resource handling to the `UuidGeneratorAndResolverImpl`, such that no client has to manage a resource (set) for the `UuidGeneratorAndResolverImpl` and then provide the `save` functionality, in which we encapsulate the cleanup functionality. Then we do not even need to throw an exception, as discussed in your other remark.

jGleitz: Yes, that sounds like a much better solution! :+1:

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: Resolved by creating the resource in a dedicated resource set within the `UuidGeneratorAndResolverImpl`.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
HeikoKlare: I have added a commit that adopts the proposal. It replaces the `cleanupRemovedElements` by a `save` method and hides the cleanup functionality in that method (currently assuming that this is the right point in time to perform the cleanup), and it creates the UUID resource within the `UuidGeneratorAndResolverImpl` instead of passing a resource to it. Do you agree with these changes, @jGleitz?

HeikoKlare:

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -310,4 +314,20 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 		}
 	}
 	
+	override save() {
+		cleanupRemovedElements()
+		uuidResource?.save(null)
```
jGleitz, line 319:
```suggestion
		uuidResource.save(null)
```

I don’t think `uuidResource` can ever be `null`, can it?

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolver.xtend
jGleitz: Yes, I do. :+1:

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: Using a resource is optional. You do not need to provide a `URI`, then no resource will be created/loaded for storing the UUIDs. In particular, views do not need a resource to store the UUID repository in, as it is transient anyway.

jGleitz: Ah, I see.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -55,48 +58,49 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
 	 * Instantiates a UUID generator and resolver with no parent resolver, 
 	 * the given {@link ResourceSet} for resolving objects
-	 * and the given {@link Resource} for storing the mapping.
+	 * and a resource at the given {@link URI} for storing the mapping in.
 	 * @param resourceSet -
 	 * 		the {@link ResourceSet} to load model elements from, may not be null
-	 * @param uuidResource -
-	 * 		the {@link Resource} to store the mapping in, may be null
+	 * @param resourceUri -
+	 * 		the {@link URI} to place a resource for storing the mapping in, may be null
```
jGleitz, line 65: I think we should document the behaviour for `null`:
```suggestion
	 * 		the {@link URI} to place a resource for storing the mapping in. If {@code null}, the 
	 * 		UUID repository will not be persisted.
```

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -55,48 +58,49 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
-	 * and the given {@link Resource} for storing the mapping.
+	 * and a resource at the given {@link URI} for storing the mapping in.
 	 * @param parentUuidResolver -
 	 * 		the parent {@link UuidResolver} used to resolve UUID if this contains no appropriate mapping, may be null
 	 * @param resourceSet -
 	 * 		the {@link ResourceSet} to load model elements from, may not be null
-	 * @param uuidResource -
-	 * 		the {@link Resource} to store the mapping in, may be null
+	 * @param resourceUri -
+	 * 		the {@link URI} to place a resource for storing the mapping in, may be null
```
jGleitz, line 81:
```suggestion
I think we should document the behaviour for `null`:
```suggestion
	 * 		the {@link URI} to persist this repository to when {@link #save} is called. If {@code null}, this 
	 * 		UUID repository will not be persisted.
```
```

File: tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend [hunk 6]
```diff
@@ -134,4 +136,62 @@ class UuidGeneratorAndResolverImplTest {
...
+		]
+		val uuid = uuidGeneratorAndResolver.generateUuid(root)
+		uuidGeneratorAndResolver.save()
+		assertEquals(uuid, uuidGeneratorAndResolver.getUuid(root))
+		resource.contents.clear
+		assertEquals(uuid, uuidGeneratorAndResolver.getUuid(root))
+		uuidGeneratorAndResolver.save()
+		assertThrows(IllegalStateException)[uuidGeneratorAndResolver.getUuid(root)]
+	}
+
```
jGleitz, line 196: Can you add a test case that saves the repository to disk and reloads it from disk, verifying that all expected UUIDs are still there?

HeikoKlare: Thanks for that question! I have added according test cases, and, of course, they revealed that reloading UUIDs from their persistence does not work properly (as the referenced objects are either proxies or erroneously resolved in the resolver's resource set). I have fixed that and provided according test cases.
As a nice side effect, this makes any storage of model instances of the `VirtualModel` obsolete, as the `UuidGeneratorAndResolverImpl` resolves all of them on load. I have to admit that we may question whether such a side effect is wanted, but I think it is valid and I have added according documentation in the `UuidGeneratorAndResolverImpl`.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -88,21 +92,27 @@ class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
+				if (resolvedObject.eIsProxy) {
+					throw new IllegalStateException("Object " + proxyEntry.key + " has a UUID but could not be resolved")
+				}
+				registerEObject(proxyEntry.value, resolvedObject)
 			}
 		}
-		return repository
+		if (uuidResource !== null) {
+			uuidResource.contents += repository
+		}
```
jGleitz, line 114: I think these three lines should not be part of this method, but rather part of the constructor. Simply because I cannot reasonably expect this assignment to occur based on the method name ‘`resolveObjectsInLoadedRepository`’.

tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend
jGleitz: I am glad to hear that more testing improved the code. That’s why we do it, after all.

> As a nice side effect, this makes any storage of model instances of the VirtualModel obsolete, as the `UuidGeneratorAndResolverImpl` resolves all of them on load. I have to admit that we may question whether such a side effect is wanted

I think it’s a nice simplification that we don’t need to save the loaded VSUM instances separately, but can do this through the UUID repository. With this simplification, we need to store less data to disk and avoid a source of inconsistency.

_However_, I think it’s worth considering whether we can make this effect clearer in the code. Currently, the VSUM models are loaded as a side effect of creating a `UuidGeneratorAndResolverImpl`. I think that’s bad for two reasons: First, developers cannot reasonably expect that, second, constructors should be cheap and free of side effects, from my point of view.

I think it’s clear that the logic for loading model instances from a UUID repository should reside in `UuidGeneratorAndResolverImpl`, because we want to hide the information about how we serialize there. So I’d propose to have a static method `loadModelInstancesFromSerializedUuidRepository` that we call _before_ creating a new `UuidGeneratorAndResolverImpl`. This is a little more costly because the constructor of `UuidGeneratorAndResolverImpl` will re-resolve the loaded model instances, but I don’t expect that cost to be high (everything is in memory already) and I think it’s worth paying it for having clearer code.

By the way, have you tried whether you can load the repository in the VSUM resource set and then move it to the UUID resource set? I’d expect this to work because proxies can be resolved there. It would also simplify the code.

HeikoKlare: I agree that it is worth considering and discussing how to separate the logic for loading the models from the `UuidGeneratorAndResolverImpl` instantiation. I made the same considerations yesterday, and it would be great to agree on some solution.

> So I’d propose to have a static method `loadModelInstancesFromSerializedUuidRepository` that we call before creating a new `UuidGeneratorAndResolverImpl`.

I am not sure whether I completely understand that proposal. If I understand correctly, one would call this method with the models' resource set and a UUID resource URI (just like the constructor), which then loads the resource and resolves all the models in the models' resource set. Is that correct?
My proposal would be to do it the other way round: We provide an instance method `loadUuidsAndModelsFromSerializedUuidRepository` (or the like) that is called _after_ creating a `UuidGeneratorAndResolverImpl`, which basically only separated the load step from the instantiation step. Then you always start with an empty resolver and can load a repository into it, which then also resolves the models. The benefit is that the constructor does not depend on the size of _any_ models anymore, as also the UUID repository loading and relinking of referenced objects is done in an explicit method then. Additionally, there is no implicit protocol to be followed for loading UUIDs (first calling a static method, then instantiating a repository for the proper resource set). The protocol is then correctly realized by the `UuidGeneratorAndResovlerImpl`. Finally, it avoid loading the repository twice (once for loading the models and once for instantiating the `UuidGeneratorAndResolverImpl`). Am I missing any drawback in that solution?

> By the way, have you tried whether you can load the repository in the VSUM resource set and then move it to the UUID resource set?

I would expect that to work, but this conflicts with our goal of separating the models' resource sets by domains (see #440). Then, there would not be a single resource set to load the UUID repository in so that it can resolve all referenced objects. That's why I did not consider that solution.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: True. In fact, I missed to push the last commit which even made the situation worse, because these lines accidentially add a second root element. I'll fix that.

tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend
HeikoKlare: We could also place the protocol for creation and loading into some factory to relieve the client from knowing it, so that he only needs to call some `createAndLoad` method. We could then also move the bunch of constructors to according factory methods.

jGleitz: That’s another possibility, yes. The logic should then go into the existing `VirtualModelBuilder`, because from my point of view, it is part of creating a virtual model.

This structure would make a lot of sense, actually: I can use a UuidResolverAndGeneratorImpl without a virtual model, but then I don’t need the loading logic. So having the loading logic in the virtual model builder makes sense. However, I still think that to have proper information hiding, the logic for actually reading the serialized repository should not leave `UuidResolverAndGeneratorImpl`.

HeikoKlare: Basically you are right, but this will be a larger enhancement than only extending the `VirtualModelBuilder`. First, loading is not only mixed up with object instantiation in the `UuidGeneratorAndResolverImpl` but also in the `ResourceRepositoryImpl` as a whole and the used `InternalCorrespondenceModelImpl`. This requires a larger refactoring to achieve that a `VirtualModel` is first instantiated with its meta configuration (i.e., the domains, the consistency specification etc.; in other words, the _V-SUMM_) and then potentially filled with existing information at the model level (model instances, UUIDs, correspondences; in other words, the _V-SUM_). This is a two-step process, which we potentially also need to represent as such.

For that reason, I propose a commit that separates `UuidGeneratorAndResolverImpl` instantation and loading existing resources as well as provides a factory for that. This still results in performing the load in the `ResourceResourceImpl` instantiation instead of the `VirtualModelBuilder` (or the like), but, as mentioned above, this is a general issue of current `VirtualModel` loading. This is why I have added issue #444 for that to be resolved completely and explicitly instead of mixing it up with the concerns of this PR.

>  the logic for actually reading the serialized repository should not leave `UuidResolverAndGeneratorImpl`.

Absolutely.

HeikoKlare: Do you agree with the proposed changes for the moment, @jGleitz?
I am already working on a proper implementation of reloading for all (model level) resources of a `VirtualModel`, which then correctly loads UUIDs, models and correspondences, to be proposed in a dedicated PR.

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -14,89 +14,69 @@ import org.eclipse.emf.ecore.EClass
...
-		if (repository === null) {
-			repository = UuidFactory.eINSTANCE.createUuidToEObjectRepository
-			if (uuidResource !== null) {
-				uuidResource.contents += repository
+	/**
+	 * Loads the mapping between UUIDs and {@link EObject}s from the persistence at the {@link URI}
+	 * given in the constructor and resolves the referenced {@link EObject}s in the {@link ResourceSet}
+	 * given in the constructor.
+	 */
+	def loadUuidsAndModelsFromSerializedUuidRepository() {
```
jGleitz, line 65:
```suggestion
	def package loadUuidsAndModelsFromSerializedUuidRepository() {
```

This should only be called by the Factory. Visibility is one reason why Factories are often nested classes (e.g. `UuidGeneratorAndResolverImpl.Factory`):  that gives them access to private function, without the need to open the visibility higher than necessary.

I’m fine with having `package` visibility, although I like the cleaner visibilities with the nested class pattern. `public` is too much.

tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend
jGleitz: Since we have #444, I’m fine with merging the current state (except for the method visibility issue I commented).

bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
HeikoKlare: Ah, that's true. In fact, the complete `UuidGeneratorAndResolverImpl` has `package` visibility, so this will actually not make a difference, but it clarifies who should have access to that. Having a factory as a nested class has the drawback that you need to expose the class for which you provide a factory. Here, we hide the implementation class and only provide the interface, so I think the structure is fine as it is.

tests/framework/tools.vitruv.framework.uuid.tests/src/tools/vitruv/framework/uuid/tests/UuidGeneratorAndResolverImplTest.xtend
HeikoKlare: Alright, thanks! I've committed the visibility suggestion. 
And yes, I am also not completely convinced of the currently state, but it only makes sense to improve it together with the other load functionalities. We can then proceed with the discussion on how to properly design that in an according PR for that.

PR #445, opened 2021-03-03
Add GitHub Action for Validation Against Applications

### LINKED ISSUE #427: GitHub Actions Check for Validation Against Applications
I just want to paste a tought and would be interested in your opinions.
We know that large parts of the framework are not tested (well) and that we have to run our CBS applications to implicitly test framework changes. We should, of course, add (unit/integration) tests for the framework, at least for regression when fixing bugs. We will, however, not be able to provide sophisticated tests that make validation against the CBS applications obsolete soon.

Currently, I execute the CBS applications against framework changes locally. This requires building the framework and the applications (and potentially the domains) against the local updatesite generated by the framework build. This is cumbersome and slows down the development process due to manual effort and usage of local system resources for that validation.
This process is something we could also provide as an additional GitHub Actions check for the framework. We could also agree on a naming convention for branches of the domains and applications projects that adapt to changes of the framework to test them all together.

Although I know that this will neither replace proper framework tests nor be pretty efficient (such a check will potentially run for more than half an hour), I think it could improve feedback for changes.
As mentioned before, I would be interested in your opinions.

HeikoKlare:
This PR adds a GitHub action that validates the [CBS applications](https://github.com/vitruv-tools/Vitruv-Applications-ComponentBasedSystems) against the framework changes propsed in a pull request. See the discussion in #427.

It uses the `master` branches of the CBS domains and applications or, if existing, the branches with the same name as the one of the pull request. This will currently only work if the branch is present in the repository of the `vitruv-tools` organization and not in a fork.

Closes #427.

### DIFF
--- a/.github/workflows/applicationsvalidation.yml
+++ b/.github/workflows/applicationsvalidation.yml
@@ -1,0 +1,92 @@
+name: Application Validation
+
+on:
+  pull_request:
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Framework
+        uses: actions/checkout@v2
+        with:
+          path: framework
+      - name: Checkout CBS Domains
+        uses: actions/checkout@v2
+        with:
+          path: cbsdomains
+          repository: vitruv-tools/Vitruv-Domains-ComponentBasedSystems
+          ref: master
+      - name: Checkout CBS Applications
+        uses: actions/checkout@v2
+        with:
+          path: cbsapplications
+          repository: vitruv-tools/Vitruv-Applications-ComponentBasedSystems
+          ref: master
+      - name: Checkout Matching Domains/Applications Branches
+        run: |
+          cd cbsdomains
+          git checkout -B ${{ github.head_ref }}
+          cd ../cbsapplications
+          git checkout -B ${{ github.head_ref }}
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml', '**/MANIFEST.MF') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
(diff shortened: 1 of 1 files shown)


jGleitz:
Currently, the pipeline does not transfer the built artefacts from the build to this step. To do this, I think the job should either be in the main CI pipeline (to share the same file system) or transfer the built artefacts through GitHub Actions’ artefact action.

.github/workflows/applicationsvalidation.yml
```diff
@@ -0,0 +1,92 @@
...
+          sed -i 's#https://vitruv-tools.github.io/updatesite/nightly/framework/#file:///${maven.multiModuleProjectDirectory}/../framework/releng/tools.vitruv.updatesite.aggregated/target/final#wresult' cbsapplications/releng/tools.vitruv.applications.cbs.parent/pom.xml
+          if [ ! -s result ]; then exit 1; fi;
+          sed -i 's#https://vitruv-tools.github.io/updatesite/nightly/domains/cbs/#file:///${maven.multiModuleProjectDirectory}/../cbsdomains/releng/tools.vitruv.domains.cbs.updatesite.aggregated/target/final#wresult' cbsapplications/releng/tools.vitruv.applications.cbs.parent/pom.xml
+          if [ ! -s result ]; then exit 1; fi;
+      - name: Build Framework
+        uses: GabrielBB/xvfb-action@v1
+        with:
+          working-directory: ./framework
+          run: >
+            ./mvnw -B clean verify
```
jGleitz, line 55: I’d propose to only run `package` here and set the appropriate options to skip all tests. This will speed up the job and tests will be run by the other job anyway.

jGleitz: Plus, it might be interesting to see whether changes are compatible with the applications even if the unit tests still fail.

HeikoKlare: Ah, that's a great idea, thanks! I already merged to use the validation for further PRs as soon as possible, but I will of course provide an update that implements that improvement.

PR #450, opened 2021-03-04
Do not Execute Tests in GH Action for Application Validation

HeikoKlare:
Only runs `package` instead of `verify` for the framework projects in application validation GitHub action.
See proposal in #445.

In addition, fixes the checkout of branches from the domains and applications repositories, which was not working properly before.

### DIFF
--- a/.github/workflows/applicationsvalidation.yml
+++ b/.github/workflows/applicationsvalidation.yml
@@ -41,7 +41,7 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 11
-      - name: Adapt Updatesite
+      - name: Adapt Updatesites
         run: |
           sed -i 's#https://vitruv-tools.github.io/updatesite/nightly/framework/#file:///${maven.multiModuleProjectDirectory}/../framework/releng/tools.vitruv.updatesite.aggregated/target/final#wresult' cbsdomains/releng/tools.vitruv.domains.cbs.parent/pom.xml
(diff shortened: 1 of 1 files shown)


.github/workflows/applicationsvalidation.yml
```diff
@@ -52,7 +54,7 @@ jobs:
         with:
           working-directory: ./framework
           run: >
-            ./mvnw -B clean verify 
+            ./mvnw -B package
```
jGleitz, line 57: Currently, tests are still executed. I don’t know whether the below also works for tycho-surefire

```suggestion
            ./mvnw -B package -Dmaven.test.skip=true
```

HeikoKlare: If I understand correctly, only Maven surefire tests are executed (i.e., tests in projects with the `.tests-without-platform` marker), as the plugin runs in the `test` phase, which is executed for the `package` goal. However, the Tycho surefire tests do not seem to executed, as they are run in the `integration-test` phase. Thus you proposal should be fine to also disable the Maven surefire tests.

jGleitz:
Skipping tests seem to save 3m30s :+1:

HeikoKlare:
Yes 🎉 that what a great idea! 
Having automated validation against the applicationss in less than 20 minutes is much faster than I expected, and I think it's really worth it to get some feedback on compatibility of changes.

PR #452, opened 2021-03-12
VURI Simplification and Extension

HeikoKlare:
* Simplifies the existing VURI logic and enables it to handle arbitrary URIs (file URIs were not supported before), also removing special case handling for file-based URIs
* Extends VURIs to support EObjects, reusing resolution logic from `UuidGeneratorAndResolverImpl`
* Make `UuidGeneratorAndResolverImpl` use central VURI logic for getting uniformly resolvable EObject URI

In consequence, this PR especially centralizes the logic for determining resolvable URIs for EObjects in VURI rather than the `UuidGeneratorAndResolverImpl` without exposing the logic.

### DIFF
--- a/bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
+++ b/bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
@@ -8,9 +8,12 @@ import java.util.Map;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.emf.common.CommonPlugin;
 import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.emf.ecore.util.EcoreUtil;
 
 import tools.vitruv.framework.util.bridges.EMFBridge;
+import static com.google.common.base.Preconditions.checkState;
 
 /**
  * Implements the multiton design pattern.
@@ -19,95 +22,140 @@ import tools.vitruv.framework.util.bridges.EMFBridge;
  *
  */
 public class VURI implements Comparable<VURI> {
-    private static final Map<String, VURI> INSTANCES = new HashMap<String, VURI>();
+	private static final Map<String, VURI> INSTANCES = new HashMap<String, VURI>();
 
-    private org.eclipse.emf.common.util.URI emfURI;
+	private URI emfURI;
 
-    /** Multiton classes should not have a public or default constructor. */
-    private VURI(final String uriString) {
-        this.emfURI = EMFBridge.createURI(uriString);
-    }
+	/** Multiton classes should not have a public or default constructor. */
+	private VURI(final String uriString, final URI uri) {
+		this.emfURI = uri;
+	}
 
-    public static synchronized VURI getInstance(final String key) {
-        VURI instance = INSTANCES.get(key);
-        if (instance == null) {
-            instance = new VURI(key);
-            String newKey = instance.toString();
-            VURI oldInstance = INSTANCES.get(newKey);
-            if (oldInstance != null) {
-                INSTANCES.put(key, oldInstance);
-                return oldInstance;
-            } else {
-                // we also have to map the actual string representation of the key after the vuri
-                // was
-                // created because a prefix may be prepended to the key while the VURI is created
-                INSTANCES.put(newKey, instance);
-            }
-            INSTANCES.put(key, instance);
-        }
-        return instance;
-    }
+	private static synchronized VURI putInstance(final String key, final URI uri) {
+		checkState(!INSTANCES.containsKey(key), "already contains %s", key);
+		VURI newInstance = new VURI(key, uri);
+		VURI existingInstance = INSTANCES.get(newInstance.toString());
+		if (existingInstance != null) {
+			checkState(existingInstance.emfURI.equals(uri), "%s was already mapped to URI '%s' while adding '%s'", key,
+					existingInstance.emfURI, uri);
+			INSTANCES.put(key, existingInstance);
+			return existingInstance;
+		} else {
+			INSTANCES.put(key, newInstance);
+			// We also have to map the actual string representation of the key after the
+			// VURI was created because a prefix may be prepended to the key while the VURI
+			// is created
+			INSTANCES.put(newInstance.toString(), newInstance);
+			return newInstance;
+		}
+	}
 
-    public static VURI getInstance(final Resource resource) {
-        return getInstance(resource.getURI());
-    }
+	public static synchronized VURI getInstance(final String key) {
+		VURI instance = INSTANCES.get(key);
+		if (instance == null) {
+			instance = putInstance(key, EMFBridge.createURI(key));
+		}
+		return instance;
+	}
 
-    public static VURI getInstance(final URI uri) {
-        checkNotNull(uri, "URI for VURI");
-        if (null == uri.toFileString()) {
-            return getInstance(uri.toString());
-        }
-        return getInstance(uri.toFileString());
-    }
+	public static VURI getInstance(final Resource resource) {
+		return getInstance(resource.getURI());
+	}
 
-    public static VURI getInstance(final IResource iResource) {
-        String[] keyStrSegments = iResource.getFullPath().segments();
-        StringBuilder keyString = new StringBuilder();
-        for (int i = 0; i < keyStrSegments.length; i++) {
-            if (i > 0) {
-                keyString.append("/");
-            }
-            keyString.append(keyStrSegments[i]);
-        }
+	public static VURI getInstance(final URI uri) {
+		checkNotNull(uri, "URI for VURI");
+		if (null == INSTANCES.get(uri.toString())) {
+			putInstance(uri.toString(), uri);
+		}
+		return getInstance(uri.toString());
+	}
 
-        return getInstance(keyString.toString());
-    }
+	public static VURI getInstance(final IResource iResource) {
+		String[] keyStrSegments = iResource.getFullPath().segments();
+		StringBuilder keyString = new StringBuilder();
+		for (int i = 0; i < keyStrSegments.length; i++) {
+			if (i > 0) {
+				keyString.append("/");
+			}
+			keyString.append(keyStrSegments[i]);
+		}
 
-    @Override
-    public String toString() {
-        return this.emfURI.toString();
-    }
+		return getInstance(keyString.toString());
+	}
 
-    public String toResolvedAbsolutePath() {
-        return CommonPlugin.resolve(this.emfURI).toFileString();
-    }
+	public static VURI getInstance(EObject object) {
+		return getInstance(getResolvableUri(object));
+	}
 
-    public URI getEMFUri() {
-        return this.emfURI;
-    }
+	private static URI getResolvableUri(EObject object) {
+		// we cannot simply use EcoreUtil#getURI, because object’s domain might use XMI
+		// UUIDs. Since XMI UUIDs can be different for different resource sets, we
+		// cannot use URIs with XMI UUIDs to identify across resource sets. Hence, we
+		// force hierarchical URIs. This assumes that the resolved object’s graph has
+		// the same topology in the resolving resource set. This assumption holds when
+		// we use this method.
+		Resource resource = object.eResource();
+		if (resource == null) {
+			return URI.createURI(EcoreUtil.getRelativeURIFragmentPath(null, object));
+		} else {
+			int rootElementIndex = 0;
+			EObject resourceRoot;
+			if (resource.getContents().size() <= 1) {
+				resourceRoot = object.eResource().getContents().get(0);
+			} else {
+				// move up containment hierarchy until some container is one of the resource's
+				// root elements
+				EObject container = object;
+				while (container != null && (rootElementIndex = resource.getContents().indexOf(container)) == -1) {
+					container = container.eContainer();
+				}
+				checkState(container != null, "some container of %s must be a root element of its resource", object);
+				resourceRoot = container;
+			}
+			String fragmentPath = EcoreUtil.getRelativeURIFragmentPath(resourceRoot, object);
+			if (fragmentPath == "") {
+				return resource.getURI().appendFragment("/" + rootElementIndex);
+			} else {
+				return resource.getURI().appendFragment("/" + rootElementIndex + "/" + fragmentPath);
+			}
+		}
+	}
 
-    public String getFileExtension() {
-        return this.emfURI.fileExtension();
-    }
+	@Override
+	public String toString() {
+		return this.emfURI.toString();
+	}
 
-    public String getLastSegment() {
-        String lastSegment = this.emfURI.lastSegment();
-        return (lastSegment == null ? "" : lastSegment);
-    }
+	public String toResolvedAbsolutePath() {
+		return CommonPlugin.resolve(this.emfURI).toFileString();
+	}
 
-    /**
-     * Returns a new VURI that is created from the actual VURI by replacing its file extension with
-     * newFileExt
-     *
-     * @param newFileExt
-     * @return the new VURI with the replaced file extension
-     */
-    public VURI replaceFileExtension(final String newFileExt) {
-        return VURI.getInstance(this.emfURI.trimFileExtension().appendFileExtension(newFileExt));
-    }
+	public URI getEMFUri() {
+		return this.emfURI;
+	}
 
-    @Override
-    public int compareTo(final VURI otherVURI) {
-        return this.emfURI.toString().compareTo(otherVURI.toString());
-    }
+	public String getFileExtension() {
+		return this.emfURI.fileExtension();
+	}
+
+	public String getLastSegment() {
+		String lastSegment = this.emfURI.lastSegment();
+		return (lastSegment == null ? "" : lastSegment);
+	}
+
+	/**
+	 * Returns a new VURI that is created from the actual VURI by replacing its file
+	 * extension with newFileExt
+	 *
+	 * @param newFileExt
+	 * @return the new VURI with the replaced file extension
+	 */
+	public VURI replaceFileExtension(final String newFileExt) {
+		return VURI.getInstance(this.emfURI.trimFileExtension().appendFileExtension(newFileExt));
+	}
+
+	@Override
+	public int compareTo(final VURI otherVURI) {
+		return this.emfURI.toString().compareTo(otherVURI.toString());
+	}
 }
--- a/bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/VitruviusConstants.java
+++ b/bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/VitruviusConstants.java
@@ -6,6 +6,7 @@ public class VitruviusConstants {
 	private static final String EXTENSION_PRIORITY_PROPERTY_NAME = "priority";
 	private static final String PATHMAP_PREFIX = "pathmap:/";
 	private static final String PLATFORM_RESOURCE_PREFIX = "platform:/resource/";
+	private static final String FILE_PREFIX = "file:/";
 	private static final String CORRESPONDENCES_FILE_EXT = "correspondence";
 	private static final String UUID_FILE_EXT = "uuid";
 	private static final String DEFAULT_NAME_OF_IDENTIFIER_FEATURE = "id";
@@ -31,6 +32,10 @@ public class VitruviusConstants {
 
 	public static String getPlatformResourcePrefix() {
 		return PLATFORM_RESOURCE_PREFIX;
+	}
+	
+	public static String getFilePrefix() {
+		return FILE_PREFIX;
 	}
 
 	public static String getPathmapPrefix() {
--- a/bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/bridges/EMFBridge.java
+++ b/bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/bridges/EMFBridge.java
@@ -85,7 +85,8 @@ public final class EMFBridge {
 	 */
 	public static URI createURI(String uriString) {
 		if (uriString != null) {
-			if (!uriString.startsWith(VitruviusConstants.getPlatformResourcePrefix())
+			if (!uriString.startsWith(VitruviusConstants.getFilePrefix())
+					&& !uriString.startsWith(VitruviusConstants.getPlatformResourcePrefix())
 					&& !uriString.startsWith(VitruviusConstants.getPathmapPrefix())) {
 				if (startsWithWindowsDriveLetterColonBackslash(uriString) || uriString.startsWith("/")) {
 					return URI.createFileURI(uriString);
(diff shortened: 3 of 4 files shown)


bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
```diff
@@ -19,95 +22,140 @@
...
+		return getInstance(getResolvableUri(object));
+	}
+
+	private static URI getResolvableUri(EObject object) {
+		// we cannot simply use EcoreUtil#getURI, because object’s domain might use XMI
+		// UUIDs. Since XMI UUIDs can be different for different resource sets, we
+		// cannot use URIs with XMI UUIDs to identify across resource sets. Hence, we
+		// force hierarchical URIs. This assumes that the resolved object’s graph has
+		// the same topology in the resolving resource set. This assumption holds when
+		// we use this method.
```
jGleitz, line 96: This comment does not make sense in the new context.

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
```diff
@@ -19,95 +22,140 @@
...
+		}
+
+		return getInstance(keyString.toString());
+	}
+
+	public static VURI getInstance(EObject object) {
+		return getInstance(getResolvableUri(object));
+	}
+
+	private static URI getResolvableUri(EObject object) {
```
jGleitz, line 90: The contract of this method is very specific to the use case we have in `UuidGeneratorAndResolverImpl`. I don’t find it very easy to understand why it needs to be the way it is if I don’t have the specific context from `UuidGeneratorAndResolverImpl`. As a test, try to specify the contexts in which it is valid to use such a VURI, and the contexts in which it is invalid (and one should rather use a UUID, for example).

Additionally, I think the URI generated by this method is not guaranteed to work. As I understand it, resources don’t _have_ to support this method of addressing objects. It’s just that all we have encountered do.

For this reason, I’d generally advise against putting this logic somewhere where it might get used by others. What is your use case for this?

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
```diff
@@ -19,95 +22,140 @@
...
-     */
-    public VURI replaceFileExtension(final String newFileExt) {
-        return VURI.getInstance(this.emfURI.trimFileExtension().appendFileExtension(newFileExt));
-    }
-
-    @Override
-    public int compareTo(final VURI otherVURI) {
-        return this.emfURI.toString().compareTo(otherVURI.toString());
-    }
+	private static final Map<String, VURI> INSTANCES = new HashMap<String, VURI>();
```
jGleitz, line 25: While we are on it: Have we ever investigated whether this cache improves anything? Especially since it’s unbounded, it might grow quite large. 

Note also that:
 * EMF’s URI apply a (bounded) object pool for URI objects, which makes this class a light wrapper around pooled objects
 * VURI instances are often short-lived in code I’ve seen, which makes them good candidates for fast GC. 
 * Because the cache uses string keys, it cannot save us the computations for determining the VURI (because caching happens afterwards)
 * The cache forces getting a VURI to be `synchronized`, which can have a severe performance impact.


From my point of view, the cache is likely to cause more harm than good, and we should remove it if we don’t have good arguments for keeping it.

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
```diff
@@ -19,95 +22,140 @@
  *
  */
 public class VURI implements Comparable<VURI> {
```
jGleitz, line 24: We might consider removing this class altogether and making all methods utility methods for EMF’s URI. I currently don’t see good reasons to keep VURI instances around. As far as I can see, they don’t abstract away anything.

Note that VURIs do, in particular, not abstract from EMF’s URIs. Because while clients don’t need to reference the EMF URI type, using VURIs is inherently tied to EMF’s URI behaviour.

jGleitz: My reasoning for removing this class is that if somebody asked me what a `VURI` is, and how it is different from an EMF URI, I would have no good answer. I’d probably say: It’s like an EMF URI, only that Vitruvius prefers to use its own type.

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
HeikoKlare: Actually, the logic is not specific to the `UuidGeneratorAndResolverImpl` but specific to the problem that we have resources loaded multiple times and need to match their contents. In fact, that is the reason why we have these IDs at all, as they bridge between different identities of the same objects. In consequence, everywhere we need to bridge this gap, we have to use such an URI resolution mechanism, which is currently encapsulated into UUIDs.

Unfortunately, UUIDs appear to be insufficient for what we want to in Vitruv: When we only store original changes and expect the consequential ones to be reproducible by applying consistency preservation, this only work if consistency preservation produced elements with the exact same UUIDs, as later original changes to the elements produced by consistency preservation need to depend on the UUID. Thus, their benefit reduces to ease tracking of elements while they are not contained anywhere, which was previously done by a cache in the TUID mechanism.
Thus, to proceed with implementing necessary functionality in the framework, we need to overcome UUIDs, probably resulting in some kind of more lightweight TUIDs. Resolvable URIs may be sufficient as a generic way of calculating them.

However, this affects large parts of the framework and I have not thought all the process and its consequence through. But to ease this, a first step is to remove UUIDs from the correspondence model, as correspondences can actually reference objects without any drawback (if I do not miss anything).
The only drawback is that you can then only resolve correspondences for objects of the resource sets of the `VirtualModel`, as the indirection using UUIDs does not bridge this gap anymore. Since the correspondence model should be internal to the `VirtualModel`, this is conceptually not a problem. But currently many tests access the correspondence model, although they should only access a _view to_ the correspondence model, if at all. Therefore, we need to bridge the gap between resources multiple times also for correspondences rather than only model elements having a UUID.

All in all, this is only a minor step in a potentially long process to finally support what we actually want to support in Vitruv. Maybe the VURIs will then be integrated into some other representation that replaces UUIDs and VURIs, but that's not yet clear to me, which is why I'll try to incrementally approach a solution.

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
HeikoKlare: I also had the intention to completely remove that class some time ago. I'll figure out whether that's possible. But you are right that there should actually be no necessity for it.

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
HeikoKlare: This is not a cache. The actual intention was that multiple URIs map to the same VURI. E.g., you can reference a resource by its file, platform or even pathmap URI, but all of them point to the same VURI. I know that this was necessary some year ago (already before I digged into Vitruv), but I am not sure whether this is still necessary. I could image that in the UI code there may be some necessity, but resources are accessed in different ways. But I am neither sure that this is the case, nor that mapping the URIs to the same VURI then actually resolved the problem reliably.

jGleitz: >  The actual intention was that multiple URIs map to the same VURI

But why does it have to be the _same_ VURI? I can see use cases where multiple URIs should be mapped to an _equal_ VURI, but I don’t see a good reason why it should need to be _the same_.

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
jGleitz: > When we only store original changes and expect the consequential ones to be reproducible by applying consistency preservation

I didn’t know that this was a use case. If I understand you correctly, you want to apply the same original changes multiple times and reach the same UUIDs each time. When do we need that? Is that part of making Vitruv version-control-ready?

If the only objective is to gain the same UUIDs for the same change sequence, we can have that a lot cheaper. We could simply make calculation UUIDs a function of the existing UUIDs. So that alone doesn’t seem like a good reason to abandon the current UUID approach. It is not clear to me that, in fact, ‘we need to overcome UUIDs’.


---

> But to ease this, a first step is to remove UUIDs from the correspondence model

It might be reasonable to make the correspondence model reference objects instead of UUIDs, or it might not be. I can’t judge. However, I don’t see the connection between referencing objects in the correspondence model and the changes we are discussing here. If we can

> only resolve correspondences for objects of the resource sets of the VirtualModel

we can just use normal URIs (which pontentially contain XMI IDs) in the serialization, can’t we?

---

To summarize, I think I have two issues:

First, you seem to suggest moving from UUIDs to a mechanism that is conceptually close to TUIDs, only with a different syntax. Note that I am _generally in favour_: From my point of view, having the Virtual Model store yet another kind of metadata (the UUIDs) is a major drawback. And I am still not convinced that the TUID approach cannot work (at least if we have proper Java change monitoring). _However_, I am suspicious of your approach to ‘slowly approach’ this change. From my point of view, we have just recently gained a somewhat stable and clean mechanism for identifying objects with the removal of TUIDs and the cleanup of the UUID logic. Moving away from that, without a clear plan, seems unwise to me. (I will not have to shoulder the consequences of any of these decisions, but I still feel inclined to share my view on them.

Second, I cannot see the connection between the change to how we identify objects and the changes you propose in this PR. As I explained above, I think we should not move the ‘build a special URI’ logic out of `UuidGeneratorAndResolverImpl’, and I have not yet understood how what you are planning to do makes the move necessary.

HeikoKlare: > If I understand you correctly, you want to apply the same original changes multiple times and reach the same UUIDs each time. When do we need that? Is that part of making Vitruv version-control-ready?

In the first place, yes. Conceptually, a V-SUM should only store original changes. Any system state can then be derived by applying them (together with consistency preservation), just like in an ordinary version control system. And just like there, one wants to branch and merge changes, which, in fact, must always be a rebase operation when using inductive consistency preservation, during which consistency preservation is executed anew. Thus, you cannot rely on making UUID generation a function depending on the existing UUIDs, as you propose.

> And I am still not convinced that the TUID approach cannot work (at least if we have proper Java change monitoring).

Of course, it can. We only removed the existing TUID mechanism because it provided so much code for functionality that is not needed that it was not worth adapting, while the UUID mechanism was there and fitted to our current needs.

> Moving away from that, without a clear plan, seems unwise to me.

Well, I never said that there is no plan. I just said that I am not yet sure how the final design will look like, because I am not able to forsee all the pitfalls that we need to consider for a proper realization. I do not see a problem in moving towards that solution with reasonable smaller changes in an agile way. To be honest, I don't get the point why incrementally performing changes (that are themselves reasonable) is worse than having one big bang replacement that no one completely understands.
To be able to replace or adapt UUIDs so that they support the intended use case, we need to do some further cleanup first (like removing the indirection between correspondences and objects), which is reasonable already by itself.

> we can just use normal URIs (which pontentially contain XMI IDs) in the serialization, can’t we?

Resolving corresponding objects means that you find correspondences containing one set of objects and returning the corresponding one. If the given set of objects is from a different resource set, resolving them by their URIs results in the same problem of mismatching XMI IDs like in the UUID resolution mechanism.

Finally, to come back to the contents of this PR: I still don't see why determining URIs that allow to resolve an object in different instances of the same resource is something that is tied to UUIDs. We did not use it somewhere else before, but, as I mentioned in the previous post, it's actually about briding this gap of resolving objects in different instances of the same resource, which is not necessarily tied to UUIDs.

jGleitz: > To be honest, I don't get the point why incrementally performing changes (that are themselves reasonable) is worse than having one big bang replacement that no one completely understands.

Well that’s the crux: I don’t think that the changes are reasonable. You are, of course, right, that incremental development is usually better that the ‘big bang’ change. But what I am trying to say is this: I see reasons why we shouldn’t make the URI logic more public, so I think there must be convincing reasons to do it anyway. Especially in the first formulation, your plan sounded to be like ‘let’s pull this out, use it, and see where it leads us’. And I objected based on that impression.

> Resolving corresponding objects means that you find correspondences containing one set of objects and returning the corresponding one.

But you cannot use the hierarchical URIs of this function, either, can you? They can only be used if the different resource sets are guaranteed to be in the same state. That is not the case when resolving corresponding objects, is it? We might try to resolve a correspondence for an object that has been moved in the view, while it has not been moved in the VSUM yet.

This limitation is exactly why I am so hesitant to make this logic more public. It does _not_ ‘allow to resolve an object in different instances of the same resource’. Instead, it only allows to do this if both instances are guaranteed to be in the same state. This precondition, however, is not visible from the code and very easy to break. By extension, the logic that we make more public will be easy to misuse.

HeikoKlare: I don't know why we need to discuss about the unclear phrasing I used in previous comments. I have already clarified that. To make it even more precise: We still suffer from the problem that large parts of our applications (more precisely the application tests) rely on some internals of the `VirtualModel`. In this case, they rely on the ability to resolve correspondences in the `VirtualModel` resource set while only having access to model elements in their view resource set. There are two perfect solutions, which are, first, to change their implementation to not rely on correspondences and, second, to provide a view mechanism that allows the creation of composed views including correspondences (which would also require a large refactoring of the application tests).

Since I meanwhile work on completely removing the VURIs, the proposed change probably degrades to an explicit `getFragmentBasedURI` or the like anyway, which can easily make potential restrictions explicit. In addition, we can have it marked as an internal API, such that using it will be discouraged and the application tests get according warnings without the necessity to update them completely before having a proper view mechanism that requires them to be changed again.

We can, of course, find other workarounds to keep the tests working until they become refactored. However, I still don't see why the logic of calculating fragment-based URIs should be specific to UUIDs. In partciular because
> They can only be used if the different resource sets are guaranteed to be in the same state

applies to hierachic URIs in general and not only to those we provide. You can only avoid this by having URIs based on UUIDs stored _within_ the model elements. But that depends on the individual metamodel and not the mechanism of determining URIs.
This is why I do not see why your last argument should be valid, as it applies to EMFs URIs in general and not only to the ones we use.

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
HeikoKlare: To be honest, I don't know. From the code I've seen I would guess that it was supposed to avoid equivalence checking and only require reference equality checks. But I am not sure whether mapping the different URIs to the same VURIs is still necessary anywhere. At least, I did not find any code that uses this functionality yet when removing the map (and VURIs in general).

bundles/framework/tools.vitruv.framework.util/src/tools/vitruv/framework/util/datatypes/VURI.java
jGleitz: I don’t think you have replied to this objection:

> you cannot use the hierarchical URIs of this function for [resolving correspondences], either, can you? They can only be used if the different resource sets are guaranteed to be in the same state. That is not the case when resolving corresponding objects, is it?

From what I was able to understand from our discussion so far, is that you want to expose the fragment URI logic to use it to resolve correspondences. My objection is that if fear that this would not lead to correct results.

---

> The proposed change probably degrades to an explicit getFragmentBasedURI or the like anyway, which can easily make potential restrictions explicit

Exposing a method called `getFragmentBasedURI` is definitely better, as it communicates the restrictions. That would not have been the case if we just had an API that was ‘give me a URI for this object’.

However, as detailed above, I still fear that actually using the URIs elsewhere is not correct. 

---

> don't know why we need to discuss about the unclear phrasing I used in previous comments

We don’t. My intention was to show you why my first objection was very strong and general, and that my stance has changed since, although I now object for different, more specific reasons. There is obviously a lot of misunderstanding going on between us in this thread, and I hoped making transparent where I’ve misunderstood you previously would help with that.

jGleitz: >  Resolving correspondences from the correspondence model with objects from another resource set leads to the UuidGeneratorAndResolverImpl resolving the objects within the resource set of the VirtualModel by applying the URI calculation method we are talking about. 

As far as I understand Vitruv, this is just not true: Resolving correspondences is based on UUIDs, and UUIDs are assigned as soon as the model objects are created or loaded. Hence, using a fragment URI is always correct, as long as resources are never loaded while there is unsaved in-memory state. Hence, everything works as long as the following assumption is true:

 >  Models are loaded from disk only at ‘save points’, where there is no unsaved state.

If we use segment URIs to resolve correspondences, however, we get a much stricter assumption:

 > Correspondences are only resolved across resource sets at ‘save points’, where there is no unsaved state.

I don’t think that these assumptions are ‘equally bad’, I think the second is stronger and, hence, worse.

---

> This mechanism only works by accident

I wouldn’t frame it that way: The mechanism works because it relies on a mechanism that is designed to allow resolving across resource sets. However, I agree, of course, with the later statement

> which should not be necessary for objects resolved in the corresponcence model as it should only be accessed within the `VirtualModel` using `EObjects` residing in the `VirtualModel` resource set anyway.

---

My proposal: 

 * regarding this method: I think exposing internally it with the name you propose is fine
 * regarding correspondences: Do not support resolving across resource sets in the normal API, but rather build an additional API that explicitly allows resolving across resource sets. Document that that API should not be used, similar to `LegacyVitruvApplicationTest`.

HeikoKlare: > Resolving correspondences is based on UUIDs, and UUIDs are assigned as soon as the model objects are created or loaded. Hence, using a fragment URI is always correct, as long as resources are never loaded while there is unsaved in-memory state. 

If I am not completely mistaken, this is only true as long as you use the `UuidResolver` in which the object is registered. However, in this case the element is not registered in that resolver. Calling `CorrespondenceModel.getCorrespondingObjects(#[object])` with `object` not being in the resource set used by the `UuidResolver` referenced in the correspondence model (which, in our case, is the one of the `VirtualModel`) leads to using the resource-set-spanning resolution mechanism of the `UuidGeneratorAndResolverImpl`. So calling that method with `object` from a view resource set (like is done in the test) will _not_ resolve the element by its UUID but by its URI.
In consequence, both implementations have equal assumptions.

Thanks for your proposal! Of course, resolving across resource sets will not be part of the normal API. Probably it would be best to adapt the test cases at least so far that they do not access any instance of `CorrespondenceModel` anymore (whether it is the `VirtualModel`-internal one or a view) but only call some test API method that completely hides the logic for resolving corresponding objects in whatever way it is realized.

jGleitz: > leads to using the resource-set-spanning resolution mechanism of the `UuidGeneratorAndResolverImpl`

I don’t know what you mean by that [there is no such mechanism when resolving a UUID](https://github.com/vitruv-tools/Vitruv/blob/bb537f93bbf047066876adb9e355136d779c757f/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend#L141). My understanding is that there should be a `UuidGeneratorAndResolverImpl` in the resource set of the test view. This resolver has the VSUM resolver as its parent. As soon as we load resources in the test view, the UUIDs get loaded from the VSUM using the `loadUuidsIntoChild` mechanism. Hence, they should be assigned when loading the objects, not when resolving the correspondence. This is why I think I described the assumptions accurately.

After writing this, I realized that we [have a resource set spanning mechanism for the other direction](https://github.com/vitruv-tools/Vitruv/blob/bb537f93bbf047066876adb9e355136d779c757f/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend#L111) However, I don’t understand why we need it.

HeikoKlare: > Hence, they should be assigned when loading the objects, not when resolving the correspondence.

When you pass an object from your test view's resource set to the correspondence model, how should it access the view's `UuidGeneratorAndResolver` to retrieve its UUIDs? Of course, the view's `UuidGeneratorAndResolver` knows the mapping between the UUIDs and the objects, as it is instantiated during resource loading, but not the one within the `VirtualModel` to which the correspondence model has access. This one only knows the models of the resource set within the `VirtualModel`.

> After writing this, I realized that we have a resource set spanning mechanism for the other direction However, I don’t understand why we need it.

I am not sure whether we still need it. It was probably required for the Java workarounds, because the objects within changes passed to the `VirtualModel` resided within the view's resource set(s) and not within the one of the `VirtualModel`. Maybe this functionality is obsolete by now (and by accident only keeps the correspondence resolution in the test cases working).

jGleitz: >  how should it access the view's UuidGeneratorAndResolver to retrieve its UUIDs

Ah, that’s where I went wrong. Okay, I think now I get it. The process is like this:
 
 1. The view passes a view object to the correspondence model.
 2. The correspondence model queries the UUID of that object _using [the resource set spanning mechanism](https://github.com/vitruv-tools/Vitruv/blob/bb537f93bbf047066876adb9e355136d779c757f/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend#L111)_ (so yes, we still need it, exactly for this case)
 3. The correspondence model looks up the correspondence by ID
 4. The correspondence model queries the vsum object by UUID
 5. The correspondence model returns the vsum object
 6. Optionally: the test view resolves the vsum object to a test view object.
 
 So yes, you are right: The mechanism relies already on the stronger assumption and your proposal makes nothing worse.
 
 We should test whether we can remove [the resource set spanning mechanism](https://github.com/vitruv-tools/Vitruv/blob/bb537f93bbf047066876adb9e355136d779c757f/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend#L111) from `internalGetUuid`, because as we have discussed here, it relies on bad assumptions.

HeikoKlare: Exactly. I'm sorry: giving you such a structured process description would have made it easier for you to understand what I wanted to explain.

> We should test whether we can remove the resource set spanning mechanism from internalGetUuid, because as we have discussed here, it relies on bad assumptions.

Totally agree. Then a valuable result of our discussion (in addition to that agree w.r.t this PR) is that we can remove some obsolete and potentially error-hiding functionality and simplify the UUID resolver implementation :-)

PR #453, opened 2021-03-16
State-Based Propagation Improvement

### LINKED ISSUE #269: State-based propagation of moved resources
The operation `propagateChangedState(Resource newState, URI oldLocation)` is experimental by now (i.e. it was not tested yet).
We have to add some regression tests to figure out if the method operates as expected and keeps doing so in the future.

HeikoKlare:
This PR fixes and improves the state-based change propagation. It prepares for state-based propagation of changes in Java by the Java code monitor.
* Fixes essential bugs in the default state-based propagation strategy preventing Java state comparison
* Adds support for creating and deleting resources rather than only for their modification and extends the API accordingly
* Properly extracts the information passed to the `VirtualModel` to perform state-based change propagation
* Adds a bunch of basic test cases for the state-based change propagation

⚠️ depends on #455 to ensure that UUIDs for readonly resources work properly for state comparison.

Closes #269

### DIFF
--- a/bundles/framework/tools.vitruv.framework.domains/src/tools/vitruv/framework/domains/DefaultStateBasedChangeResolutionStrategy.xtend
+++ b/bundles/framework/tools.vitruv.framework.domains/src/tools/vitruv/framework/domains/DefaultStateBasedChangeResolutionStrategy.xtend
@@ -38,10 +38,6 @@ class DefaultStateBasedChangeResolutionStrategy implements StateBasedChangeResol
 		changeFactory = VitruviusChangeFactory.instance
 	}
 	
-	private def createResourceSet() {
-		new ResourceSetImpl().awareOfDomains(domainRepository)
-	}
-	
 	private def checkNoProxies(Resource resource, String stateNotice) {
 		val proxies = resource.referencedProxies
 		checkArgument(proxies.empty, "%s '%s' should not contain proxies, but contains the following: %s", stateNotice, resource.URI, String.join(", ", proxies.map[toString]))
@@ -52,13 +48,15 @@ class DefaultStateBasedChangeResolutionStrategy implements StateBasedChangeResol
 		checkArgument(oldState !== null && newState !== null, "old state or new state must not be null!")
 		newState.checkNoProxies("new state")
 		oldState.checkNoProxies("old state")
-		val resourceSet = createResourceSet()
-		val currentStateCopy = oldState.copyInto(resourceSet)
+		val oldResourceSet = new ResourceSetImpl()
+		val newResourceSet = new ResourceSetImpl()
+		val currentStateCopy = oldState.copyInto(oldResourceSet)
+		val newStateCopy = newState.copyInto(newResourceSet)
 		val diffs = currentStateCopy.record(resolver) [
 			if (oldState.URI != newState.URI) {
-				currentStateCopy.URI = newState.URI
+				currentStateCopy.URI = newStateCopy.URI
 			}
-			compareStatesAndReplayChanges(newState, currentStateCopy)
+			compareStatesAndReplayChanges(newStateCopy, currentStateCopy)
 		]
 		return changeFactory.createCompositeChange(diffs)
 	}
@@ -69,7 +67,7 @@ class DefaultStateBasedChangeResolutionStrategy implements StateBasedChangeResol
 		newState.checkNoProxies("new state")
 		// It is possible that root elements are automatically generated during resource creation (e.g., Java packages).
 		// Thus, we create the resource and then monitor the re-insertion of the elements
-		val resourceSet = createResourceSet()
+		val resourceSet = new ResourceSetImpl().awareOfDomains(domainRepository)
 		val newResource = resourceSet.createResource(newState.URI)
 		newResource.contents.clear()
 		val diffs = newResource.record(resolver) [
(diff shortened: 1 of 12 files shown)


bundles/framework/tools.vitruv.framework.domains/src/tools/vitruv/framework/domains/DefaultStateBasedChangeResolutionStrategy.xtend
```diff
@@ -27,55 +27,87 @@ import static tools.vitruv.framework.uuid.UuidGeneratorAndResolverFactory.create
...
-			return changeRecorder.changes
-		}
+	private def compareStatesAndReplayChanges(Notifier newState, Notifier currentState) {
+		val scope = new DefaultComparisonScope(newState, currentState, null)
+		val comparison = EMFCompare.builder.build.compare(scope)
+		// Assign the eResource of a root element, as otherwise the DomainAwareResource is used
+		// and can lead to a mismatch with the ordinars resource in the EMF merger 
+		comparison.matchedResources.forEach[
+			it.right = it.right.contents.get(0).eResource()
+			it.left = it.left.contents.get(0).eResource()
```
JanWittler, line 104: If there is no better alternative like customising the EMF merger to be aware of `DomainAwareResource`, I think we should expose some getter to the underlying resource of a `DomainAwareResource` instead of requiring this workaround. The code makes assumptions which may change in the future and is not understandable at all if not already worked with `DomainAwareResource` before.

JanWittler (changes requested):

tests/framework/tools.vitruv.framework.domains.tests/src/tools/vitruv/framework/tests/domains/BasicStateChangePropagationTest.xtend
```diff
@@ -0,0 +1,272 @@
...
+				contents += root => [
+					id = "Root"
+				]
+			] >> modelResource
+		]
+		EcoreResourceBridge.saveResource(-modelResource)
+
+		resourceSet.record [
+			root.id = "Change"
+		]
```
JanWittler, line 103: I am wondering whether changing an element's ID is a reasonable scenario for which we should have a test. From my understanding, the ID is static and will not change during an element's lifetime. This assumption is also used by the default state based change resolution strategy as it uses EMF Compare's ID based matching (if an ID is present). Therefore, if there was some domain were the ID is likely to change, it should be configured with a custom resolution strategy where the ID based matching is disabled (and use a separate test case in that case).

bundles/framework/tools.vitruv.framework.domains/src/tools/vitruv/framework/domains/DefaultStateBasedChangeResolutionStrategy.xtend
jGleitz: I agree, the code effectively exploits a bug in the implementation of `DomainAwareResource`.

HeikoKlare: I agree that this code has to be replaced. However, conditional functionality based on the type of resource does not sound like a proper improvement. I'll try to figure out how to properly adapt that code.

tests/framework/tools.vitruv.framework.domains.tests/src/tools/vitruv/framework/tests/domains/BasicStateChangePropagationTest.xtend
HeikoKlare: That's true. The test is actually not supposed to test the ID change but to test a replacement of the complete model contents (which, for EMF Compare, is effectively the same). It then tests the proper behavior covered by the code discussed in the other comment.
I'll update the test case accordingly.

HeikoKlare: I've provided a changed of that test case explicitly focussing on replacing the root element rather than changing its ID.

bundles/framework/tools.vitruv.framework.domains/src/tools/vitruv/framework/domains/DefaultStateBasedChangeResolutionStrategy.xtend
HeikoKlare: It was even more simple to fix that: We only need a domain-aware resource set for the case that we create a new model, in which we do not apply EMF Compare anyway, because in the other cases the resource contents are only copied into the resource set, which is why domain-related load/save options are irrelevant. So I simplified the resource sets for the other cases (including the one in which EMF Compare is applied) to ordinary resource sets, which makes the workaround obsolete.

tests/framework/tools.vitruv.framework.domains.tests/src/tools/vitruv/framework/tests/domains/BasicStateChangePropagationTest.xtend
```diff
@@ -0,0 +1,274 @@
...
+						id = "ContainedRoot"
+					]
+				]
+			] >> modelResource
+		]
+		EcoreResourceBridge.saveResource(-modelResource)
+
+		resourceSet.record [
+			containedRoot.id = "Changed"
+		]
```
JanWittler, line 175: I would also change this test to not update an element's id.

JanWittler (changes requested):

tests/framework/tools.vitruv.framework.domains.tests/src/tools/vitruv/framework/tests/domains/BasicStateChangePropagationTest.xtend
HeikoKlare: Thanks for having an eye on that! I thought I checked for other ID replacements in the tests, but obviously I was not successful ;-) 
I've adapted the test case.

PR #455, opened 2021-03-18
Simplify UUID Calculation for Readonly Resources

HeikoKlare:
* Removes the UUID logic for generating UUIDs for pre-existing elements (archive and pathmap resources, such as the Java standard library or type libraries) and for storing the UUIDs for these elements. Replaces it by simply using the (immutable) URI as the UUID.
* Removes all logic related to UUIDs for pre-existing elements and their propagation to parent `UuidResolvers`
* Removes obsolete logic for resolving proxies in `UuidGeneratorAndResolverImpl`
* Requires objects for which UUIDs are generated to not be proxy elements
* Replaces logic that silently and unintentionally loads resources into the resource set of a parent `UuidResolver`

### DIFF
--- a/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
+++ b/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
@@ -24,6 +24,8 @@ import static extension tools.vitruv.framework.util.ResourceSetUtil.withGlobalFa
  */
 package class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
 	static val logger = Logger.getLogger(UuidGeneratorAndResolverImpl)
+	static val NON_READONLY_PREFIX = "ord_"
+	
 	val ResourceSet resourceSet
 	val Resource uuidResource
 	val UuidResolver parentUuidResolver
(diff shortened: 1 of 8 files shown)


bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend
```diff
@@ -138,7 +152,20 @@ package class UuidGeneratorAndResolverImpl implements UuidGeneratorAndResolver {
...
 
-	private def EObject internalGetEObject(String uuid) {
+	private def EObject getEObjectOrNull(String uuid) {
+		return uuid.getEObjectIfReadonlyUri()
+			?: uuid.getStoredEObject()
+			?: null
+	}
+	
+	private def getEObjectIfReadonlyUri(String uuid) {
+		val potentialURI = URI.createURI(uuid);
```
jGleitz, line 162: This looks expensive to me. Couldn’t we just ensure that „normal“ UUIDs have a prefix that makes them distinct?

HeikoKlare: URI creation does not seem to be that expensive. Nevertheless, you are right that we can simply add a prefix to "normal" UUIDs to avoid unnecessary overhead.
We could even add a prefix to both kinds of UUID to be absolutely sure that there will never be a change that leads to URIs of readonly resources matching the prefix of "normal" UUIDs. This will introduce additional overhead for substring calculation to retrieve the URI from a UUID of readonly resources, but I would expect those readonly elements to be referenced rarely.

HeikoKlare: I've added a prefix for "normal" UUIDs like you proposed. We will soon also change the way in which UUIDs for non-readonly elements are generated (probably also relying on ordinary URIs), such that the UUID storage will probably become obsolete anyway.

PR #467, opened 2021-04-01
Replace UUIDs with Lightweight IDs and Simplify Change and Propagation Specification Composition

### LINKED ISSUE #129: VirtualModel.propageChange with propagatedChange from other virtual model
For the versioning of **Vitruv** models, it should be possible to reapply a `PropagatedChange` recorded in one `VirtualModel` to another `VirtualModel`. This is necessary to execute a *checkout* of the `VirtualModel` like in *git*. Thus one can get an exact copy of the `VirtualModel` committed by someone else. If only the `originalChange` of the `PropagatedChange`  is applied, most likely there will be differences in the generated **UUIDs** in the `consequentialChanges`.

### LINKED ISSUE #326: Domains Should be Able to Adapt UUID Resolving Logic
[`UUIDGeneratorAndResolverImpl`](https://github.com/vitruv-tools/Vitruv/blob/master/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend) contains [logic specific to the Java domain](https://github.com/vitruv-tools/Vitruv/blob/master/bundles/framework/tools.vitruv.framework.uuid/src/tools/vitruv/framework/uuid/UuidGeneratorAndResolverImpl.xtend#L357) that tries to get around JaMoPP-quirks. This logic should be part of the Java domain.

To achieve this, one solution would be to allow a `VitruvDomain` to provide the `UUIDGeneratorAndResolver` that should be used for it. The VSUM should then delegate UUID resolving to the resolver that is responsible for the object in question.

HeikoKlare:
This PR makes significant changes to essential parts of the framework, in particular to the ID mechanism (replacing the UUID mechanism with a more lightweight one), changes and their composition, and change propagation specifications.
Most important, it replaces our UUIDs with a lightweight ID mechanism that does not depend on deriving IDs from a `VirtualModel` but allows to monitor changes completely independent from a `VirtualModel`. In addition, it enables change merging (thus is inevitable for versioning / variant management).

#### ID Mechanism
* UUIDs are replaced by IDs which are only valid at a specific point in time (comparable to TUIDs we had before).
* IDs are completely transient. A resolver can be instantiated ad-hoc and must only live for a transaction (a scope in which a removed element may be inserted again).
* ID management is completely hidden in the change projects. Only the applicators and recorders provided by the projects access the ID management (including an according `IdResolver`) but no one else has to care about IDs.
* Since IDs are completely transient, they are not persisted anymore. Instead, the `VirtualModel` simply persists the URIs of the models it manages, which were resolved implicitly from the stored IDs before.

#### Change Composition
* Complex composition hierarchies of changes are replaced by a simple composition technique only consisting of basic `TransactionalChanges` and composing `CompositeContainerChanges`. `ConcreteChange` and `CompositeTransactionalChange` has been removed.
* `TransactionalChange` encapsulates all `EChanges` relevant for one transaction (a scope after which no elements are still dangling around, i.e., have not been removed from a container to be inserted afterwards).
* `ChangeRecorder` delivers a single `TransactionalChange` instead of a list of them.

#### Change Propagation Specification Composition
* We had change propagation specifications for `VitruviusChanges`, which then, at some point, decomposed into `EChanges`, depending on which kinds of specification have been composed in which way and which order. The specifications for `EChanges` added additional functionality to apply change propagation to them. This was hard to understand and even erroneous when composing changes differently than we did before (which is why bugs occurred when changing the change composition hierarchy). 
* `ChangePropagationSpecifications` now accept `EChanges` instead of `VitruviusChanges`, because finally the specifications have to be executed for each `EChange` one after another anyway. They implement simple composition according to the composite pattern.
* The `ChangePropagator` calls the specifications for one `EChange` after another.

In addition to these specific changes, the PR removes some obsolete functionality and even completely projects to avoid their unnecessary adaptation to the above mentioned changes.

Closes #326.
Closes #129.


tsaglam (approved):
I looked through the changes and did not spot anything that stood out. Looks good to me!

Issue #425, opened 2021-02-18
Correspondence not removed after cascade deletion by container


JanWittler:
When an object is deleted as part of the deletion of its container, its correspondences are not removed.

To illustrate this, take the example of the [UmlToJavaClassMethodTest#testDeleteMethod](https://github.com/vitruv-tools/Vitruv-Applications-ComponentBasedSystems/blob/3f45c1008d9a62daede4ca5d36e7abe3362f7155/tests/tools.vitruv.applications.umljava.tests/src/tools/vitruv/applications/umljava/tests/uml2java/UmlToJavaClassMethodTest.xtend#L111). There, an UML operation with one parameter is deleted. Each the operation and the parameter have a correspondence to a java method / parameter. After the UML operation is deleted, the operation correspondence is removed. However, the parameter correspondence still exists (in [vsum]/correspondence/Correspondences.correspondence) even though the parameter does not exist anymore.

HeikoKlare:
I propose the following solution:
We perform an automated cleanup of the correspondence model (potentially during save), which removes all correspondences that relate elements that have all been removed from the models. That covers the case that complete element trees have been removed with their contained elements (like the mentioned parameters of operations).
In case there are correspondences of which only some elements have been removed from the resources but others still exist, I would consider that as an indicator for a potential mistake, because potentially the corresponding elements have not been removed by mistake. Since we only use 1:1 correspondences, this means that one element has been removed but the corresponding element still exists in the models. Thus, I would propose to log a warning or even error in that case (or should we even throw an exception as the behavior is unintended?), such that the related transformation can be fixed accordingly.

I made a prototypical implementation and if am not mistaken, most dangling correspondences are properly removed. Only in very rare cases a warning is logged, because actually a correspondence was not removed properly. I will propose this in a pull request, such that we can better figure out its consequences and merge it in case we agree that it solves the mentioned problem.

PR #469, opened 2021-04-21
Record Move from Removed Element Containment

HeikoKlare:
The `ChangeRecorder` removes the recording adapter from elements as soon as they are removed from a containment. If, however, elements contained within the removed elements (or any of its transitively contained elements) are then moved to another container, only the insertion change is recorded without recording the change for removing the element from the previous container. In consequence, applying the changes backwards for assigning appropriate IDs fails, because the elements is not inserted into its original container.
This was not recognized before as we have no situation in which we re-insert an element contained in the tree that was removed from a model containment. It first came up when re-enabling layout information in Java in #468.

This PR adapts the `ChangeRecorder` such that it defers the removal from recording to after the current transactions, such that further changes in the removed elements are still recorded. As soon as the transaction ends, elements will not be inserted again (the same assumption it made for identifying delete changes) and then the recording adapter is removed.
It also adds according regression tests.

### DIFF
--- a/bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/ChangeRecorder.xtend
+++ b/bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/ChangeRecorder.xtend
@@ -53,6 +53,7 @@ class ChangeRecorder implements AutoCloseable {
 	val IdResolver idResolver
 	val EChangeIdManager eChangeIdManager
 	val Set<EObject> existingObjects = new HashSet
+	val Set<Notifier> toDesinfect = new HashSet
 	val ResourceSet resourceSet
 	
 	new(ResourceSet resourceSet) {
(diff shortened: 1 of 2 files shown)


JanWittler:
I can image that this also caused #425 and thus will resolve it.

HeikoKlare:
Unfortunately, I think that this will not resolve #425. The changes of this PR only affect very rare cases. Only if an element that is contained in a removed one is reinserted somewhere else, there will be a difference after merging this PR. Related to the example in #425, it would only make a difference if you reinserted a parameter from a removed operation into another operation afterwards. And even then, it does not affect correspondences but only fixes a problem that was not present when we used UUIDs but came up when switching to the new ID mechanism.

But thanks to you bringing up the issue, I thought about a solution for it and wrote a proposal in #425. Maybe it's quite easy to solve, even if not implicitly resolved by this PR :-)

PR #478, opened 2021-11-05
Support for move operations

HansMartinA:


bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/resolve/EChangeResolverAndApplicator.xtend
```diff
@@ -47,7 +48,14 @@ class EChangeResolverAndApplicator {
 		val oldObject = eChange.oldContainedEObject
 		ApplyEChangeSwitch.applyEChange(eChange, forward)
 		if (eChange.isContainmentChange || affectedId != idResolver.getAndUpdateId(affectedObject)) {
-			affectedObject.updateIds(idResolver)
+			if (eChange instanceof RootEChange) {
+				val contents = eChange.resource.contents
+				for (var idx = eChange.index; idx < contents.size; idx++) {
+					updateIds(contents.get(idx), idResolver)
+				}
```
HeikoKlare, line 55: If I understand correctly, this implementation tries to solve the problem that IDs of root elements can change upon addition/removal of a root element that is not the last within the resource. I would expect this problem to occur for any kind of multi-valued feature (like the root elements of a resource conceptually are), so that we should have such an implementation for any kind of multi-valued feature.
Anyway, we need a regression test for a problematic root change as well as for multi-valued feature changes to see whether the problem occurs there as well.

HeikoKlare (changes requested):
Thanks for the improvements! We need to have some regression tests for the covered situations, and we have to investigate whether the additional ID update is necessary in more situations (at least I would expect it to be). This is, however, something we can also provide, so the requested changes do not necessarily need to be performed by you, @HansMartinA, except you require the PR to be merged soon.

JanWittler:
This PR seems to be abandoned. As there are not too many changes in here anyways, I would suggest closing the PR without merging? @HeikoKlare

HeikoKlare:
Agree

PR #479, opened 2021-12-03
Proper implementation of VirtualModel views

HeikoKlare:
This PR provides an implementation of views on a `VirtualModel`. By now, views had to be realized implicitly by loading the resources represented in the `VirtualModel`. Now the `VirtualModel` has the capability of creating views, which makes the workaround absolete and allows for more sophisticated view generations rather than one-to-one projections from the `VirtualModel`. With this PR, basic one-to-one mappings are supported but based on interfaces that allow more complex mappings.

### View Generation
A `View` is based on a `ViewType`, which describes the transformation logic for generating a view out of some kind of `ViewSource`. In our case, this `ViewSource` is realized by the `VirtualModel`. For a given `ViewSource`, a `ViewType` creates a `ViewSelector`, which allows to perform a selection of elements to be shown in the view (such as the selection of a specific Java class or of the classes to be shown in a UML diagram). After performing the selection, the `ViewSelector` is able to create a `View`.

To avoid an exposure of the `ViewSource`, a `ViewProvider` can create a `ViewSelector` for a given `ViewType`, such that handling the `ViewSource` is hidden behind that interface. In our case, the `VirtualModel` is a `ViewProvider`.

### View and View Type Implementations
With this PR, the basic implementations `BasicViewType`, `BasicModelView` and `BasicViewElementSelector` are provided. The `BasicViewType` performs a one-to-one mapping of resources within the `VirtualModel` to a view. The `BasicViewElementSelector` allows to perform a selection based on the root elements given to the `ViewType` from the `ViewSource`. The `BasicModelView` then provides a modifiable view, which records changes to given elements and allows to commit these changes to the underlying model repository.


bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+public interface View extends AutoCloseable {
+
+	/**
+	 * Provides the root model elements of this view.
+	 * 
+	 * @throws IllegalStateException if called on a closed view.
+	 * @see View#isClosed()
+	 */
+	Collection<EObject> rootObjects();
```
HeikoKlare, line 25: Should we call this `getRootObjects` or is it intentionally called `rootObjects`?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
```diff
@@ -0,0 +1,113 @@
...
+
+	/**
+	 * Specifies whether the view was modified.
+	 */
+	boolean isModified();
+
+	/**
+	 * Specifies whether the underlying virtual model has changed.
+	 */
+	boolean hasVSUMChanged();
```
HeikoKlare, line 46: Should be called `hasUnderlyingSourceChanged` or something like that, because (1) the word V-SUM is not used within the implementation and (2) views are not tied to a `VirtualModel` but to a `ViewSource`

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
```diff
@@ -0,0 +1,113 @@
...
+	 * @throws IllegalStateException if called on a closed view.
+	 * @see View#isClosed()
+	 */
+	// TODO TS: Some views may not always record changes, and thus require
+	// transactional record-and-commit.
+	List<PropagatedChange> commitChanges();
+
+	/**
+	 * Checks whether the view was closed. Closed view should not further be used.
+	 */
```
HeikoKlare, line 81: Extend documentation, becaue "should not further be used" actually means that most of the operations will throw exceptions when called after closing the view.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
```diff
@@ -0,0 +1,116 @@
...
+		val originalURI = originalResource.URI
+		val tempURI = originalURI.trimFileExtension.appendSegment("save." + originalURI.fileExtension)
+		originalResource.URI = tempURI
+		originalResource.save(null)
+		originalResource.URI = originalURI
+		val viewResource = newResourceSet.getResource(tempURI, true)
+		viewResource.URI = originalURI
+		Files.delete(Path.of(org.eclipse.emf.common.util.URI.decode(tempURI.path)))
+		EcoreUtil.resolveAll(viewResource)
+	}
```
HeikoKlare, line 75: Magic file must be replaced by a generic temporary one.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/AbstractTreeBasedViewSelector.xtend
```diff
@@ -0,0 +1,35 @@
...
+	override getChildren(EObject element) {
+		element.calculateChildren
+	}
+
+	def static Iterable<EObject> calculateChildren(EObject element) {
+		// Needs to be implemented
+		#[]
+	}
+
+}
```
HeikoKlare, line 35: We should either finalize this implementation or move it to a separate branch and put it into a subsequent PR.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/DimensionalViewSelector.xtend
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * {@linkplain ViewSelector} that offers several dimensions in which a selection can be made to select a specific view.
+ */
+interface DimensionalViewSelector extends ViewSelector {
+    def Iterable<ViewDimension> getDimensions()
+
+    def Iterable<EObject> getElementsFor(ViewDimension dimension)
+
+    def void setSelected(ViewDimension dimension, EObject element, boolean value)
+}
```
HeikoKlare, line 15: This selector is neither used nor implemented anywhere yet, so maybe we should move it to a separate branch together with the `AbstractTreeBasedViewSelector`?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/SelectionState.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Specifies whether the selection state is essentially selected or not, while
+     * abstracting from the specific selection state.
+     * 
+     * @return true if it is selected.
+     */
+    public boolean isSelected() {
+        return selected;
+    }
+}
```
HeikoKlare, line 21: Class does not seem to be used yet.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/TreeBasedViewSelector.xtend
```diff
@@ -0,0 +1,17 @@
...
+ */
+interface TreeBasedViewSelector extends ViewSelector {
+    def EObject getRoot()
+
+    def Iterable<EObject> getChildren(EObject element)
+
+	// TODO TS/TK we need to distinguish manual selection and automatic selection.
+    
+    def boolean isComplete()
+}
```
HeikoKlare, line 17: See comment for `AbstractTreeBasedViewSelector`.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/ViewDimension.xtend
```diff
@@ -0,0 +1,14 @@
...
+interface ViewDimension { // TODO TS if these signatures stay this way, extract shared interface with ViewSelector (ElementSelector/Selecting/SelectionMaking)
+
+    def Iterable<EObject> getElements()
+
+    def void setSelected(int index, boolean value)
+
+    def boolean isSelected(int index)
+
+    def Iterable<EObject> getSelectedElements()
+}
```
HeikoKlare, line 14: This seems to be a duplication of a `ModifiableViewSelection`. Do we really need it?

tests/framework/tools.vitruv.framework.vsum.tests/src/tools/vitruv/framework/tests/vsum/VirtualModelTest.xtend
```diff
@@ -42,27 +36,6 @@ class VirtualModelTest {
...
-			withChangePropagationSpecification(new RedundancyChangePropagationSpecification(aetDomain, aetDomain)).
-			withUserInteractor(
-				UserInteractionFactory.instance.createUserInteractor(
-					UserInteractionFactory.instance.createPredefinedInteractionResultProvider(null))).
-			buildAndInitialize()
-	}
-
-	private def createTestModelResourceUri(String suffix) {
-		URI.createFileURI(projectFolder.resolve("root" + suffix + ".allElementTypes").toString)
-	}
```
HeikoKlare, line 64: This method should remain and delegate to the utility class resolving the parameterized project folder rather than adding `projectFolder` all over the calls in this class.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/internal/VirtualModelImpl.xtend
```diff
@@ -203,4 +210,16 @@ class VirtualModelImpl implements InternalVirtualModel {
 		VirtualModelRegistry.instance.deregisterVirtualModel(this)
 	}
 
+	override getViewTypes() {
+		return viewTypeRepository.viewTypes
+	}
```
HeikoKlare, line 215: There is no way to fill this repository, so it will always be empty. We could either remove it, as any view type can be passed to the `VirtualModel` anyway. Or we could have the `VirtualModelBuilder` also create the view types.

HeikoKlare (commented):
I found some remaining issues to fix or discuss before merging. I did not have a look at the existing tests yet.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
HeikoKlare: Solved by placing the file for copying the UML model in an ordinary temporary file rather than creating some magic local URI.

tests/framework/tools.vitruv.framework.vsum.tests/src/tools/vitruv/framework/tests/vsum/VirtualModelTest.xtend
HeikoKlare: Fixed

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/AbstractTreeBasedViewSelector.xtend
tsaglam: I removed the unused stubs and interfaces with 4d5e5a2.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/DimensionalViewSelector.xtend
tsaglam: Same here, I removed the unused stubs and interfaces with 4d5e5a2.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/SelectionState.java
tsaglam: Same here, I removed the unused stubs and interfaces with 4d5e5a2.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/TreeBasedViewSelector.xtend
tsaglam: Same here, I removed the unused stubs and interfaces with 4d5e5a2.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selectors/ViewDimension.xtend
tsaglam: No longer needed, thus removed with 4d5e5a2.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
tsaglam: I agree, especially as it is an abbreviation. `hasUnderlyingSourceChanged` sounds good, `hasSourceChanged` would also be an option.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
tsaglam: No intention behind the name, `getRootObjects` is probably better.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
HeikoKlare: Changed the name to `haveViewSourcesChanged` in b1ff9e8. Benefit of that name: it conforms to the type `ViewSource` on which a `View` is based, hopefully improving comprehensibility.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/internal/VirtualModelImpl.xtend
HeikoKlare: Solved by providing the ability to initialize a `VirtualModel` with `ViewTypes` by adding them in the `VirtualModel Builder` in bed8efa as discussed in the ConCall.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
HeikoKlare: Fixed in c633b51.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
```diff
@@ -0,0 +1,113 @@
...
+	 * previous state and now providing a updated view on the virtual model. This
+	 * can only be done for an unmodified view.
+	 * 
+	 * @throws UnsupportedOperationException if the update is called for a dirty
+	 *                                       view.
+	 * @throws IllegalStateException         if called on a closed view.
+	 * @see View#isClosed()
+	 */
+	// TODO TS Add issue: in the long term we need to allow updates for modified
+	// views
```
tsaglam, line 59: We can remove this, I will add an issue after we merge this PR.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
```diff
@@ -0,0 +1,116 @@
...
+ */
+class BasicViewType extends AbstractViewType<BasicViewElementSelector> {
+
+	new(String name) {
+		super(name)
+	}
+
+	override createSelector(ChangeableViewSource viewSource) {
+		return new BasicViewElementSelector(this, viewSource,
+			viewSource.viewSourceModels.map[contents.head].filterNull.toList)
```
tsaglam, line 28: With `contents.head` we always take the first element of a resource's content. Do resources with multiple roots not occur here? Or do we only want to provide the first root for selection?

tsaglam (commented):
I reviewed the bundle and will review the tests separately!

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
HeikoKlare: Fixed by f8d5f68.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
```diff
@@ -0,0 +1,113 @@
...
+
+	/**
+	 * Provides all root model elements of this view that conform to a certain type.
+	 * 
+	 * @param clazz is requested root element type.
+	 * @throws IllegalStateException if called on a closed view.
+	 * @see View#isClosed()
+	 */
+	default <T> Collection<T> rootObjects(Class<T> clazz) {
+		return FluentIterable.from(rootObjects()).filter(clazz).toList();
```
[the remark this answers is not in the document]
HeikoKlare: `FluentIterable` provides a `filter` method that takes a `Class`. With an ordinary stream you have to implement an `instanceof` check.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/internal/VirtualModelImpl.xtend
```diff
@@ -203,4 +210,16 @@ class VirtualModelImpl implements InternalVirtualModel {
...
+
+	override getViewSourceModels() {
+		resourceRepository.modelResources
+	}
+
+	override <S extends ViewSelector> createSelector(ViewType<S> viewType) {
+		viewType.createSelector(this)
+	}
```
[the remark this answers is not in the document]
HeikoKlare: I have to admit that I don't get what you mean by "more complex implementation". A `ViewProvider` generates a `ViewSelector` for a given `ViewType`. In our case this is what the `VirtualModel` does, as you want to create views on a `VirtualModel`. Without this method you would simply not be able to instantiate a view.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
```diff
@@ -0,0 +1,116 @@
...
+import org.eclipse.emf.ecore.EObject
+import org.eclipse.emf.ecore.util.EcoreUtil.Copier
+import tools.vitruv.framework.vsum.views.selectors.BasicViewElementSelector
+import tools.vitruv.framework.vsum.views.ViewSelection
+import org.eclipse.emf.common.util.URI
+
+/**
+ * A basic view type that allows creating views based on a basic element-wise selection mechanism.
+ */
+class BasicViewType extends AbstractViewType<BasicViewElementSelector> {
```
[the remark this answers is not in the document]
HeikoKlare: That's true. I think the main property of the `BasicViewType` (which makes it "basic") is that it does a one-to-one mapping of resources in the view sources. Thus, we may call it something like `IdentityMappingViewType`?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
HeikoKlare: It looks like that was intended to be some prototypical implementation and now became a bug. I have provided a fix in 83e08b8. Now all root elements can be selected except for UML resources, because we have to use the copy hack for them because of `EcoreUtil.copy` not working for UML resources. Does that fit now?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/RecordingView.xtend
```diff
@@ -0,0 +1,166 @@
...
+		changeRecorder.addToRecording(viewResourceSet)
+		changeRecorder.beginRecording()
+	}
+
+	override commitChanges() {
+		checkNotClosed()
+		changeRecorder.endRecording()
+		val propagatedChanges = viewSource.propagateChange(changeRecorder.change)
+		viewChanged = false
+		update() // view shall not be dirty, thus update on commit
```
JanWittler, line 84: The documentation says
> Note that committing changes will automatically be followed by an {@link #update} from the virtual model to keep both the view and the virtual model synchronized

which reads like the Virtual Model calls the `update` method. If think it would make sense to make the Virtual Model automatically call update on any view affected by the applied changes though this will probably come with new challenges like calling update on dirty views. Thus at least the documentation should be adjusted to reflect that the `update` method should be called in the view's implementation until we can solve this better.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/RecordingView.xtend
```diff
@@ -0,0 +1,166 @@
...
+import tools.vitruv.framework.vsum.views.View
+import static com.google.common.base.Preconditions.checkArgument
+import org.eclipse.emf.common.notify.Notifier
+import org.eclipse.emf.ecore.resource.Resource
+
+/**
+ * A {@link View} that records changes to its resources and allows to propagate them 
+ * back to the underlying models using the {@link #commitChanges} method.
+ */
+class RecordingView implements ModifiableView, ChangePropagationListener {
```
JanWittler, line 27: maybe more precise: `ChangeRecordingView` ?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selection/AbstractViewSelection.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Map;
+
+import org.eclipse.emf.ecore.EObject;
+
+import tools.vitruv.framework.vsum.views.ModifiableViewSelection;
+
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public abstract class AbstractViewSelection implements ModifiableViewSelection {
```
JanWittler, line 15: I am a bit struggling with the concept of view selection. To my understanding a view selection receives a list of selectable elements and manages their selection state. Why do we then provide this option to extend view selection or provide a custom one? I.e. which scenario except for the basic yes / no selection (which is already implemented by `ElementViewSelection`) is possible that reasons to enable extension instead of providing a closed implementation? 
Or more precise: why the abstract + element classes instead of one final class to suite them all?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
```diff
@@ -0,0 +1,121 @@
...
+		val tempFilePath = Files.createTempFile(null, "." + originalURI.fileExtension)
+		val tempURI = URI.createFileURI(tempFilePath.toString)
+		originalResource.URI = tempURI
+		originalResource.save(null)
+		originalResource.URI = originalURI
+		val viewResource = newResourceSet.getResource(tempURI, true)
+		viewResource.URI = originalURI
+		Files.delete(tempFilePath)
+		EcoreUtil.resolveAll(viewResource)
+	}
```
JanWittler, line 80: We should open an issue to resolve this hack in a more elegant and maybe more general way s.t. the hack does not need to be copied for every new view type and workarounds for problems with future domains can get added easily.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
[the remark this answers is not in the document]
HeikoKlare: Done by b390fb9. I also changed `BasicViewElementSelector` to `DirectViewElementSelector` as it provides a direct selection of view elements (rather than indirectly selecting types or the like) in fe7c8b2. Let me know if you think that that's not reasonable.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/RecordingView.xtend
HeikoKlare: Yes, that's definitely more precise. Updated in f748787.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/RecordingView.xtend
HeikoKlare: I see that the documentation can be misleading, even though `{@link #update}` actually refers to the `update` method of the `View`.
There were, however, still many references to the `VirtualModel` all over the `View` interface anyway, which I corrected to references to `ViewSource` and improved the referred documentation at the same time (in 65e337d). Did I understand correctly and is it now less misleading?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/ViewType.java
```diff
@@ -0,0 +1,24 @@
...
+ * A Vitruv viewtype on the virtual model, providing a view selector and allows
+ * creating views.
+ * 
+ * @param <S> the type of view selector this view type uses
+ */
+public interface ViewType<S extends ViewSelector> {
+	/**
+	 * Returns the uniquely identifying name of the view type.
+	 */
+	String getName();
```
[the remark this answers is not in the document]
HeikoKlare: Is it not unique (and even should not be, because a `ViewType` should not know about the others). That may be a functionality of a repository but not the element itself.
I thus corrected the documentation in 64b0c16.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/ViewType.java
```diff
@@ -0,0 +1,24 @@
+package tools.vitruv.framework.vsum.views;
+
+/**
+ * A Vitruv viewtype on the virtual model, providing a view selector and allows
```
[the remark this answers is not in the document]
HeikoKlare: Corrected in 64b0c16.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
```diff
@@ -0,0 +1,114 @@
...
+	boolean isModified();
+
+	/**
+	 * Specifies whether the underlying view sources have changed.
+	 */
+	boolean haveViewSourcesChanged();
+
+	/**
+	 * Updates the view from the underlying virtual model, thus invalidating its
+	 * previous state and now providing a updated view on the virtual model. This
```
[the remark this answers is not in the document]
HeikoKlare: Corrected in 65e337d.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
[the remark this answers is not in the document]
tsaglam: Looks good!

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/internal/VirtualModelImpl.xtend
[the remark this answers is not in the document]
tsaglam: Ah, okay, now everything makes much more sense! Thanks for the explanation.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Specifies whether the view was modified.
+	 */
+	boolean isModified();
+
+	/**
+	 * Specifies whether the underlying view sources have changed.
+	 */
+	boolean haveViewSourcesChanged();
```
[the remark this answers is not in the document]
HeikoKlare: Probably, the documentation is more misleading than the method name, because it actually _is_ a question. The call does not change state but answers the question whether the view sources have changed. Thus, in my opinion, the name is correct. Renaming it to `viewSourcesHaveChanges` would be misleading as it might mean that calling it would indicate that the view sources have changed (i.e. that it changes state rather retrieving it).
My proposal would be to improve the documentation (by the way, the same applies to the method `isModified`). Do you agree?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/ModifiableViewSelection.java
```diff
@@ -0,0 +1,49 @@
...
+	 * 
+	 * @param eObject the {@link EObject} to check the selection state for
+	 * @return whether the given {@link EObject} is selected
+	 */
+	boolean isSelected(EObject eObject);
+
+	/**
+	 * Returns whether the given {@link EObject} can be selected. May only be called
+	 * for objects that are selectable (i.e., contained in
+	 * {@link #getSelectableElements()} and throws and exception otherwise.
```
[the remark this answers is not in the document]
HeikoKlare: Corrected in 64d0592.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/ModifiableViewSelection.java
```diff
@@ -0,0 +1,49 @@
...
+	 * 
+	 * @param eObject the {@link EObject} to check the selection state for
+	 * @return whether the given {@link EObject} is selected
+	 */
+	boolean isSelectable(EObject eObject);
+
+	/**
+	 * Sets the selection state of the given {@link EObject}. May only be called for
+	 * objects that are selectable (i.e., contained in
+	 * {@link #getSelectableElements()} and throws and exception otherwise.
```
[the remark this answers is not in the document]
HeikoKlare: Corrected in 64d0592.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/ViewSelector.java
```diff
@@ -0,0 +1,34 @@
+package tools.vitruv.framework.vsum.views;
+
+/**
+ * A selector for selecting the elements to be represented in a view. It
+ * encapsulates a modifiable {@link ViewSelection}, which it is able to validate
+ * and which is then passed to a created a view. It is capable of acting as a
```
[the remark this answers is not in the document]
HeikoKlare: Corrected in 64d0592.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/models/ChangeableModelRepository.java
```diff
@@ -0,0 +1,45 @@
...
+	 * 
+	 * @param newState    is the {@link Resource} of the new state
+	 * @param oldLocation specifies the previous location of the resource to avoid
+	 *                    problems with renaming or moving elements
+	 */
+	List<PropagatedChange> propagateChangedState(Resource newState, URI oldLocation);
+
+	void addChangePropagationListener(ChangePropagationListener propagationListener);
+
+	void removeChangePropagationListener(ChangePropagationListener propagationListener);
```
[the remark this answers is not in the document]
HeikoKlare: I'm not sure about the benefits of moving the functionality for a change-observable repository to a separate interface. From my point of view, observing changes in a model repository only makes sense if it can be changed, thus a change-observable model repository must always be a `ChangeableModelRepository`, i.e., they are not independent but depend on each other. You could, of course, provide a `ChangeableModelRepository` that is not observable, but the question is whether that is actually an improvement. This would, in my opinion, only be beneficial if you have several types that should be a `ChangeableModelRepository` but not observable, but if I see correctly we do actually not have that. Although I totally agree with your goal of having lean interfaces, I would rather consider this case as an improvement in complexity (due to introducting a further interface) rather than giving a benefit (such as cleaning up an interface not providing single responsibility).
But I am absolutely open for discussion. Maybe there is some reasonable principle or case-specific reason for a separation.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
tsaglam: Looks good now!

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selection/AbstractViewSelection.java
HeikoKlare: In short: Different implementations are necessary for different kinds of mappings between selectable elements and view elements. This is realized by the `isViewObjectSelected` method.

More precise: We currently only provide an `ElementViewSelection`, in which you can directly select elements of the view (such as selecting whether a Java class shall be present in the view or not). Since the selectable elements and the potential root elements of the view are the same, the `isViewObjectSelected` method implements a one-to-one mapping, e.g., if you pass a Java class to it, it returns whether the given class has been selected before.
We also plan to have further kinds of selections in the near future, in which this mapping is not one-to-one. Imagine a type selector, in which the selectable elements are types. In a concrete selector for Java, the only selectable elements may be "class" and "package". If you select "class", passing any concrete class to `isViewObjectSelected` would return `true`, but not because that specific class has been selected, but because the type "class" has been selected, which is why there is no simple one-to-one mapping like in the `ElementViewSelection`. This has to be realized by a different kind of selection. 
Since this functionality belongs to the kind of selection and not anything outside (such as the view type), it should be provided in the selection, thus requiring different implementations of `ViewSelection`.

Does that answer your question?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
HeikoKlare: Absolutely, we have to document that in an issue. If I remember correctly, we also have some functionality in the `EChange` application mechanism related to the same issue (or rather some non-functionality, because we actually accept that some changes cannot be applied although they should be, just because of the stupid implementation of the UML metamodel).
I have to admit that I expect every solution to be some kind of hack and the only elegant solution to be a fix of the UML metamodel, but maybe we can at least define some mechanism to dynamically attach specific functionality for specific kinds of resources where the ordinary functionality does work properly without implementing such domain-specific (not in the sense of a `VirtuvDomain`) functionality in the framework. But that's a discussion for the issue.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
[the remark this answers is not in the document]
HeikoKlare: Great, thanks!

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/internal/VirtualModelImpl.xtend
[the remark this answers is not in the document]
HeikoKlare: I see that it is not that obvious why we cannot simply remove that method. Do you think we need to have some further documentation so that other developers do not ask the same question?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/RecordingView.xtend
JanWittler: +1 for abstracting the virtual model away in the documentation though this was not what I meant.
The documentation reads as if the `update` method of the view that commits its changes is automatically called by another instance. However, the current solution is that we require the `commitChanges` operation to call `update` by itself. Therefore we should either adjust the documentation to something like 
> Note that committing changes should / must include calling the view's {@link #update} method ...
 
or directly provide a method like `commitChangesAndUpdate()` which just calls `commitChanges` and `update` and then discourage calling `commitChanges` directly.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/selection/AbstractViewSelection.java
JanWittler: Thanks for the explanation, totally makes sense now.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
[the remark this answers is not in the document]
JanWittler: I understand your view of thinking but honestly am still not 100% fine with the name. We could adjust the name to `areViewSourcesChanged`, then it is at least consistent with `isModified` / `isClosed` w.r.t the auxiliary verb but I don't know if it is so much better...
So I am fine with leaving it like this if no one comes up with a more precise naming alternative.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/models/ChangeableModelRepository.java
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * Propagates delta-based changes as long as the location and the name of the
+	 * resource was not changed. Should be used when there are no change sequences
+	 * available and consistency needs to be preserved based on the changes between
+	 * the previous and new state of a model.
+	 * 
+	 * @param newState is the {@link Resource} of the new state
+	 */
+	List<PropagatedChange> propagateChangedState(Resource newState);
```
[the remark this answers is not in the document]
HeikoKlare: You are right that the documentation is very confusion, in particular the way how delta-based changes were mentioned in the state-based propagated. The documentation was moved from the `VirtualModel`, which generally made comprehensibility even worse because of the changed context. Thus, have completely rewritten the documentation of the class. Can you validate that it is fine/better now?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/models/ChangeableModelRepository.java
```diff
@@ -0,0 +1,45 @@
...
+	/**
+	 * Propagates delta-based changes. Allows to change the location and the name of
+	 * the resource. Should be used when there are no change sequences available and
+	 * consistency needs to be preserved based on the changes between the previous
+	 * and new state of a model.
+	 * 
+	 * @param newState    is the {@link Resource} of the new state
+	 * @param oldLocation specifies the previous location of the resource to avoid
+	 *                    problems with renaming or moving elements
+	 */
```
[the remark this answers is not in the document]
HeikoKlare: See above. If there is still something to change, we will discuss it in the comment above.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/models/ChangeableModelRepository.java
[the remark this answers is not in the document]
JanWittler: Yes, it is way better now.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/RecordingView.xtend
HeikoKlare: I'm note sure about your proposal for adjusting the documentation, as it sounds to me like you want to have a documentation for the implementer instead of a documentation for the user (but maybe I completely misunderstand you). The documentation should tell its user what it does, not its implementer what he or she has to do. The implementer then has to see how he or she can ensure that the user-expected behavior is provided. For that reason, saying that the `update` function is called after calling `commitChanges` is fine and gives the responsibility for ensuring that functionality to the implementer of the class. So from that point of view, everything is fine in my opinion.

Despite that, your proposal to have an explicit `commitChangesAndUpdate` and reduce the existing `commitChanges` to really only commit and do not perform an update is absolutely reasonable. First (and most importantly), it will better conform to the expected behavior. Second, it can sometimes also be beneficial to have a pure `commitChanges` method that does not perform an update. For example, in the implementation of our application tests, we create a new view in each test step (for reasons of simplicity), which unnecessarily always perform an update after committing changes, although we throw away the view afterwards anyway. So as a nice side effect that may even reduce test execution times.
So, I am favor of enhancing the interface according to your proposal.

HeikoKlare: Done in 5c00bcc.

JanWittler: Just one minor point here: Even though dirty and modified can be used as synonyms, we should be consistent in the documentation and therefore keep using "modified view" instead of "dirty view" s.t. it matches the method name. Currently dirty is used in docs for `update` and `commitChanges`.

HeikoKlare: True, fixed in 1ecbb66.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/models/ChangeableModelRepository.java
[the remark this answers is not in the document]
JanWittler: My main motivation was that from an external perspective on the new changes the two methods were not really fitting into the interface contract. However, as currently only the VSUM is implementing this interface and providing the methods in question anyways, I don't see a problem to leaving the methods in the interface.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
HeikoKlare: Removed in 5631541, we just need to keep in mind that we need to open an issue.

tsaglam: I have it persisted on my TODO list.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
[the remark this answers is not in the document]
HeikoKlare: Thinking about why the auxiliary verbs are not aligned is pretty helpful. It gave me the insight that potentially the method name is pretty bad in general, because it refers to a state of the view sources rather than a state of the view, although the method if defined on a `View` rather than on a `ViewSource`. And the method is actually not about whether the view sources have changed at all, but whether they have changed since the view was last generated/updated from them. So what we actually ask when calling the method is something like `view.isUpToDateWithViewSources()`.
Do you agree and does anyone have a better name than `isUpToDateWithViewSources()`? 🙂 A short `isUpToDate()` sounds too unspecified to me.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/internal/VirtualModelImpl.xtend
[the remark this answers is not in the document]
tsaglam: I think we could add a short comment here if we want. Maybe even just a non-jdoc comment, as this mainly affects developers.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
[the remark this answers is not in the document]
tsaglam: The problem I see with `isUpToDateWithViewSources()` is that a view can technically be also "not up to date" with the source when the view is modified. However, if the view is modified this method will not return true, the method `isModified()` will.

> And the method is actually not about whether the view sources have changed at all, but whether they have changed since the view was last generated/updated from them.

But I also agree with this point.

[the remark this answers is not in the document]
JanWittler: Going one step back: Do we need exactly this method? Otherwise we could replace it with a method like `isInSyncWithViewSources()` which returns false if _either_ the view or the view sources have changed. However, then `isModified` becomes a bit ambiguous, so maybe we have to rename that too to e.g. `hasUncommittedChanges()`?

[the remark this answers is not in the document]
HeikoKlare: Right now, we do actually not need any of those methods, so from that point of view we could simply remove them. 

The use case I see for that functionality is that someone has a view and, at some point in time, wants to know whether his or her view is outdated, i.e., whether the sources have changed, such that an update must be performed (comparable to performing a `git fetch`/`git pull`). And this may be required no matter whether the view has been modified or not. Although it's good to question whether we need exactly that functionality, I would not consider some `isInSyncWithViewSources()` as a reasonable alternative, because I have no use case in mind in which this information is required. This is why I would be in favor of keeping the method semantics as it is.

If we agree on this understanding, we only have a problem finding a proper name. `haveSourcesChangedSinceLastUpdate()`  might work, but would still refer to a state of the sources rather than a state of the view, which would be an unsatisfactory result after this long discussion 😉 I would actually say that a view with modifications is still "up to date" with the sources if the sources were not modified, but I see that there can be different understandings. I do not see that problem when taking the inverse `isOutdated()`. Is there anything against that name?

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/internal/VirtualModelImpl.xtend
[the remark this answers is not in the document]
HeikoKlare: Done in 77002ae.

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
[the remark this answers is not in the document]
JanWittler: +1 for the proposed `isOutdated()` solution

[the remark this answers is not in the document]
HeikoKlare: Thanks! I have implemented the name change in 7bbdede.
@tsaglam Are you fine with the proposed name?

HeikoKlare:
If I've not missed anything, all issues/comments have been adressed (maybe except some final discussion on the method name for checking the state of a view). So after closing that discussion, I would consider this PR ready to merge. Do you agree, @tsaglam and @JanWittler ?

tsaglam (approved):
I have no further issues, so from my side, it is also ready to merge! 👍

JanWittler (approved):

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/View.java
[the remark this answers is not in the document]
tsaglam: Sounds good!

bundles/framework/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/views/impl/BasicViewType.xtend
HeikoKlare: @JanWittler Do you want to open an according issue or shall I do?

JanWittler: @HeikoKlare please do it. You know more the details why exactly this workaround is necessary for the UML domain.

HeikoKlare: Alright, I created issue #489 with an initial documentation of the problem. I also tried to explain the hack we had already implemented before for applying changes to the UML metmodels (but unfortunately not documented yet), which is a result of the same issue.

JanWittler: Perfect, thank you for the detailed explanation of the occurring issues.

PR #485, opened 2022-01-07
Disable change recording for existing and pathmap resources (#484)

### LINKED ISSUE #484: ChangeRecorder delivers changes when loading and unloading resources
### Problem
The `ChangeRecorder` can be registered for a complete `ResourceSet`, so that for every change within any resource in that resource set a change produced. In case existing resources are loaded, this leads to changes for the complete construction of this existing resource being delivered. First, this is no intended and expected behavior. Second, this leads to errors when read-only resources are processed, because the recorder tries to rollback and reapply changes, which fails on read-only resources. This is, in particular, a problem for pathmap resources of different metamodels, such as JaMoPP, which loads standard library classes as read-only pathmap resources.

The essentially same problem occurs when resources are unloaded from a recorded resource set, as changes for removing the model are generated although it is actually not removed.

Just for reasoning: I could image that this behavior was somehow intended when we used UUIDs, because then it was necessary to process every element once, even of existing resources, to generate UUIDs for them. There was then, however, also some additional logic for reloading resource to reassign the UUIDs. Since we do not use UUIDs anymore, these arguments to not apply anyway. IDs are generated ad-hoc when required, thus processing resources on loading is not required (anymore).

### Proposed Solution
To resolve this issue, the `NotificationRecorder` of the `ChangeRecorder` needs to consider whether adding and removing resource contents happens because of an actual change or because of resource loading or unloading and only produce changes in the former case.

HeikoKlare:
This PR introduces regression tests and a fix for #484.
The `NotificationRecorder` of a `ChangeRecorder` tracks whether contents are added to a resource by modification or because the resource is being loaded or unloaded and only generates changes in the former case. Before, changes were generated for resources being loaded or unloaded.

Complexity of the solution is increased because EMF does not provide a way proper way of tracking the load state of resources: The `loaded` flag is set to `true` as soon as loading is started, then notifications for adding the contents are emitted and only afterwards a notitification for setting the `loaded` flag is emitted. This makes it necessary to track the loading state of the resource in the recorder.
Vice versa, the `loaded` flag is set to `false` as soon as unloading is started and there is no simple way to distinguih between a resource being unloaded because of its deletion ad because of its removal from a resource set (which is completely different semantically). 

Fixes #484.

### DIFF
--- a/bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/ChangeRecorder.xtend
+++ b/bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/ChangeRecorder.xtend
@@ -26,6 +26,7 @@ import org.eclipse.emf.ecore.EReference
 import static com.google.common.base.Preconditions.checkNotNull
 import static com.google.common.base.Preconditions.checkArgument
 import static extension org.eclipse.emf.ecore.resource.Resource.RESOURCE__CONTENTS
+import static extension org.eclipse.emf.ecore.resource.Resource.RESOURCE__IS_LOADED
 import static extension org.eclipse.emf.ecore.resource.ResourceSet.RESOURCE_SET__RESOURCES
 import static extension tools.vitruv.framework.change.echange.resolve.EChangeResolverAndApplicator.applyForward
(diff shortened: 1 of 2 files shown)


tsaglam (approved):
Looks good to me!

PR #487, opened 2022-01-25
Add insurancePersons reactions and test cases

tintengrun:
Added reactions for insurance2persons and persons2insurance directions, with corresponding test cases


bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/insurance2persons/InsuranceToPersons.reactions
```diff
@@ -13,8 +17,125 @@ reaction CreatedInsuranceDatabase {
...
+	call createPerson(newValue)
+}
+
+routine createPerson(insurance::InsuranceClient client) {
+	match{
+		val personRegister = retrieve persons::PersonRegister corresponding to client.eContainer
+	}
+	action{
+		call {
+			if (client.gender == Gender.MALE) {
```
HeikoKlare, line 51: Please use a `switch` statement to switch over enum literal values.

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/insurance2persons/InsuranceToPersons.reactions
```diff
@@ -13,8 +17,125 @@ reaction CreatedInsuranceDatabase {
...
+	}
+	action{
+		call {
+			if (client.gender == Gender.MALE && !(person instanceof Male)) {
+				deletePerson(client)
+				createMalePerson(client)
+			} else if (client.gender == Gender.FEMALE && !(person instanceof Female)) {
+				deletePerson(client)
+				createFemalePerson(client)
+			}
```
HeikoKlare, line 123: I would propose to have a `switch` over the genders here as well (to ensure that all literals a matched) and then validate whether the person's gender does not correspond to the client's gender within the two `case` statements.

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/persons2insurance/PersonsToInsurance.reactions
```diff
@@ -13,8 +16,82 @@ reaction CreatedPersonRegister {
...
+}
+
+reaction CreatedPerson {
+	after element persons::Person created and inserted in persons::PersonRegister[persons]
+	call createInsuranceClient(newValue)
+}
+
+routine createInsuranceClient(persons::Person person) {
+	match{
+		val insuranceDatabase = retrieve insurance::InsuranceDatabase corresponding to person.eContainer
```
HeikoKlare, line 47: Same comment as for the other direction:
This implicitly assumes `eContainer` to be a `PersonRegister`, which may fail in case the metamodel changes and persons can be contained somewhere else. I would propose to have a utility method that returns the `PersonRegister` for a personand handles that case that the container is no `PersonRegister` explicitly, e.g., by throwing an exception.

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/persons2insurance/PersonsToInsurance.reactions
```diff
@@ -13,8 +16,82 @@ reaction CreatedPersonRegister {
...
+	}
+	action{
+		val newClient = create insurance::InsuranceClient and initialize {
+			newClient.name = person.fullName
+			//newClient.socialSecurityNumber = 000
```
HeikoKlare, line 52: Remove this line instead of commenting it out.

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/persons2insurance/PersonsToInsurance.reactions
```diff
@@ -13,8 +16,82 @@ reaction CreatedPersonRegister {
...
+			if (person instanceof Male) {
+				newClient.gender = Gender.MALE
+			} else {
+				newClient.gender = Gender.FEMALE
+			}
```
HeikoKlare, line 57: I would propose to have a `switch` for the type that matches all types of `Person` and also has default behavior for the case that `Person` is neither `Male` nor `Female` as result of potential metamodel change.

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
```diff
@@ -2,32 +2,445 @@ package tools.vitruv.applications.demo.insurancepersons.tests.insurance2persons
...
+import static org.hamcrest.CoreMatchers.*
+import static org.hamcrest.MatcherAssert.assertThat
+import static org.junit.jupiter.api.Assertions.assertEquals
+import static tools.vitruv.testutils.matchers.ModelMatchers.*
 
 class InsurancePersonsTest extends VitruvApplicationTest {
-	static val PERSONS_MODEL = DomainUtil.getModelFileName('model/model', new PersonsDomainProvider)
-	static val INSURANCE_MODEL = DomainUtil.getModelFileName('model/model', new InsuranceDomainProvider)
+	static val logger = Logger.getLogger(InsurancePersonsTest)
+	TestInfo testInfo = null
```
HeikoKlare, line 26: It would be sufficient to only store the test naem instead of the complete `TestInfo` object and directly access it instead of extracting the test method name in every test case.

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
```diff
@@ -2,32 +2,445 @@ package tools.vitruv.applications.demo.insurancepersons.tests.insurance2persons
...
-	def void createRoot() {
-		resourceAt(INSURANCE_MODEL).propagate[contents += InsuranceFactory.eINSTANCE.createInsuranceDatabase]
+	def void insertRegister(TestInfo testInfo) {
+		this.testInfo = testInfo
+		resourceAt(INSURANCE_MODEL).propagate[contents += InsuranceFactory.eINSTANCE.createInsuranceDatabase()]
+		assertThat(resourceAt(PERSONS_MODEL), exists);
+		assertEquals(1, resourceAt(PERSONS_MODEL).contents.size);
+		assertEquals(1, resourceAt(PERSONS_MODEL).allContents.size);
+		assertThat(resourceAt(PERSONS_MODEL).contents.get(0), instanceOf(PersonRegister));
+		assertEquals(0, resourceAt(PERSONS_MODEL).contents.get(0).eAllContents().size);
```
HeikoKlare, line 51: You may use `head` instead of `get(0)` (applies to all usages of `get(0)`. The only (maybe also useful) difference is that `head` returns `null` in case there is no element while `get(0)` throws an exception.

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
```diff
@@ -2,32 +2,445 @@ package tools.vitruv.applications.demo.insurancepersons.tests.insurance2persons
...
+		]
 	}
+	
+	def void testExample() {
+		val String surroundingMethodName = this.testInfo.getDisplayName()
+		logger.trace(surroundingMethodName + " - begin")
+		logger.trace(surroundingMethodName + " - preparation done")
+		logger.trace(surroundingMethodName + " - propagation done")
+		logger.trace(surroundingMethodName + " - finished without errors")
+	}
```
HeikoKlare, line 94: This test case should not exist.

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
```diff
@@ -2,32 +2,445 @@ package tools.vitruv.applications.demo.insurancepersons.tests.insurance2persons
...
+	
 
 	@Test
 	def void testCreateInsuranceDatabase() {
+		val String surroundingMethodName = this.testInfo.getDisplayName()
+		logger.trace(surroundingMethodName + " - begin")
+		assertThat(resourceAt(INSURANCE_MODEL), exists)
 		assertThat(resourceAt(PERSONS_MODEL), exists)
+		//TODO
```
HeikoKlare, line 103: I guess here should be a validation of the insurance database and person register instead of a "TODO".

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
```diff
@@ -2,32 +2,445 @@ package tools.vitruv.applications.demo.insurancepersons.tests.insurance2persons
...
+		
+		logger.trace(surroundingMethodName + " - preparation done")
+		
+		InsuranceDatabase.from(INSURANCE_MODEL).propagate [
+			insuranceclient += InsuranceFactory.eINSTANCE.createInsuranceClient => [name = MALE_NAME gender = Gender.MALE]
+			insuranceclient += InsuranceFactory.eINSTANCE.createInsuranceClient => [name = FEMALE_NAME gender = Gender.FEMALE]
+		]
+		InsuranceDatabase.from(INSURANCE_MODEL).propagate [
+			insuranceclient += InsuranceFactory.eINSTANCE.createInsuranceClient => [name = MALE_NAME_2 gender = Gender.MALE]
+			insuranceclient += InsuranceFactory.eINSTANCE.createInsuranceClient => [name = FEMALE_NAME_2 gender = Gender.FEMALE]
```
HeikoKlare, line 161: Format the code so that each assignment is in a single line to improve readability (like it is done below).

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
```diff
@@ -2,32 +2,445 @@ package tools.vitruv.applications.demo.insurancepersons.tests.insurance2persons
...
+		logger.trace(surroundingMethodName + " - preparation done")
+		
+		InsuranceDatabase.from(INSURANCE_MODEL).propagate [
+			val searchedClient = insuranceclient.findFirst[x|x.gender.equals(Gender.MALE)]
+			searchedClient.gender = Gender.FEMALE
+		]
+		InsuranceDatabase.from(INSURANCE_MODEL).propagate [
+			val searchedClient = insuranceclient.findLast[x|x.gender.equals(Gender.FEMALE)]
+			searchedClient.gender = Gender.MALE
+		]
```
HeikoKlare, line 261: Why have two changes of the gender instead of one here?

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
```diff
@@ -2,32 +2,445 @@ package tools.vitruv.applications.demo.insurancepersons.tests.insurance2persons
...
+			persons += PersonsFactory.eINSTANCE.createFemale => [fullName = FEMALE_NAME_2]
+		]
+		
+		assertCorrectInsuranceDatabase(expectedInsuranceDatabase)
+		assertCorrectPersonRegister(expectedPersonRegister)
+		logger.trace(surroundingMethodName + " - finished without errors")
+	}
+	
+	@Test
+	def void testChangedName() {
```
HeikoKlare, line 200: It may also make sense to have test cases for corner cases, for example setting an empty name. The current implementation should handle that the same way as non-empty names are handled, but it might be useful as a regression test.

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
```diff
@@ -2,32 +2,445 @@ package tools.vitruv.applications.demo.insurancepersons.tests.insurance2persons
...
+		assertThat(castedInsuranceDatabase, equalsDeeply(expectedInsuranceDatabase))
+	}
+	
+	def void createClientsAndPropagate() {
+		InsuranceDatabase.from(INSURANCE_MODEL).propagate [
+			insuranceclient += InsuranceFactory.eINSTANCE.createInsuranceClient => [name = MALE_NAME gender = Gender.MALE]
+			insuranceclient += InsuranceFactory.eINSTANCE.createInsuranceClient => [name = FEMALE_NAME gender = Gender.FEMALE]
+			insuranceclient += InsuranceFactory.eINSTANCE.createInsuranceClient => [name = MALE_NAME_2 gender = Gender.MALE]
+			insuranceclient += InsuranceFactory.eINSTANCE.createInsuranceClient => [name = FEMALE_NAME_2 gender = Gender.FEMALE]
+		]
```
HeikoKlare, line 85: Please remove this method an create exactly those elements required in a specific test case. Currently this is used as method to generate the highest number of clients required in any of the test, but someone reading the tests must always take a look at this method to comprehend how the model is actually initialized.
In case there is more complex setup logic shared between multiple tests (e.g., creating one male and one female person), you may separate that in a helper method. But then at least the names of the clients should be explicitly passed to that method, so that is is not necessary to read the helper when reading a test case to understand it (and where it possibly fails).

HeikoKlare (changes requested):
I have made some rather detailed comments for improvements of the Reactions and some more general comments for the improvement of the test cases. The comments regarding the `InsuranceToPersonsTest` apply in the same way to the `PersonsToInsuranceTest`, I just did not want to repeat them.

Some further general comments that do not apply to single lines of code:
- Code formatting: there are some empty lines, missing whitespaces (in particular before blocks, starting with `{` and false indentations. Please clean up that formatting.
- Code comments: there are some commented out statements (especially `socialSecurityNumber = 000`). Please remove them.
- The test cases have `logger.trace` statements for some kind of lifecycle logging. Are they necessary for anything specific? Otherwise I would be in favor of removing them, because they reduce comprehensibility of the code and will probably not give a benefit, as a test case failure will tell you where it failed anyway.

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/insurance2persons/InsuranceToPersons.reactions
tintengrun: What should I put in the `default`-case? My first idea was an IllegalArgumentException

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
tintengrun: To check both the male to female as well as female to male change

HeikoKlare: Okay, then I have two remarks:
1. If both kinds of changes shall be validated, the model state between these two changes should be checked as well.
2. If I understand correctly, with these two changes the resulting model looks exactly like the original model, so if the Reactions do not support gender changes, the test would still succeed. This problem would implicitly be resolved by checking the state between the two changes as proposed in 1. as well.

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/insurance2persons/InsuranceToPersons.reactions
HeikoKlare: If you define a case for each enum literal, there is no essential necessity to have a `default` case. In case another enum literal will be added in the future, the compiler will give a warning. If you want to add a `default` case though, throwing an `IllegalArgumentException` would be fine.

tintengrun: Fixed, with an exception in the default case. See commit 921b7cb5639471ec829b754d64432b2b8a878f63

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/insurance2persons/InsuranceToPersons.reactions
tintengrun: Fixed as suggested in 921b7cb5639471ec829b754d64432b2b8a878f63

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/persons2insurance/PersonsToInsurance.reactions
tintengrun: Removed in 921b7cb5639471ec829b754d64432b2b8a878f63

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/persons2insurance/PersonsToInsurance.reactions
tintengrun: Fixed in eca4f96fe288d427e735ec480969d57536646d87

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
tintengrun: Removed all manual logging from test cases in insurancepersons. See 921b7cb5639471ec829b754d64432b2b8a878f63

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
tintengrun: Fixed in 921b7cb5639471ec829b754d64432b2b8a878f63

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
tintengrun: Removed test case in 921b7cb5639471ec829b754d64432b2b8a878f63

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
tintengrun: Fixed in 921b7cb5639471ec829b754d64432b2b8a878f63

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
tintengrun: Added separate test cases for both m2f and f2m changes, so only one person changes at a time. See bcaa2b3d4dbc241ad68e8896dd93cf1600e3a483

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
tintengrun: Removed method and instead create all clients where they are needed. See 921b7cb5639471ec829b754d64432b2b8a878f63

bundles/applications/tools.vitruv.applications.demo.insurancepersons/src/tools/vitruv/applications/demo/insurancepersons/insurance2persons/InsuranceToPersons.reactions
```diff
@@ -13,8 +17,125 @@ reaction CreatedInsuranceDatabase {
...
+			} else {
+				createFemalePerson(client)
+			}
+		}
+	}
+}
+
+routine createMalePerson(insurance::InsuranceClient client) {
+	match{
+		val personRegister = retrieve persons::PersonRegister corresponding to client.eContainer
```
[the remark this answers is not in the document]
tintengrun: Added helper method in 0104fb8dc6f07e78b411e1bfae5ba2c51bb1d785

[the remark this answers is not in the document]
HeikoKlare: There were still some `eContainer` references left, which I replaced in d1b8a67.

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
HeikoKlare: Now each test case creates the same clients as the auxiliary method created before. However, most test cases do not need to have four clients but less are sufficient, often even one or none. Please reduce the test initialization to what is necessary for the test case to keep it as simple/minimal as possible.

HeikoKlare (changes requested):
The changes are fine now except for the initialization in the `InsuranceToPersonsTest` cases, which in many cases is overly complex (see detailed comment).

tests/applications/tools.vitruv.applications.demo.insurancepersons.tests/src/tools/vitruv/applications/demo/insurancepersons/tests/insurance2persons/InsurancePersonsTest.xtend
tintengrun: Slimmed down test cases in acb4a598b3d6173bee36f01bef3e0f8d39b15de1. I left in some test objects that do not change in the test cases where I believe they might be affected if something went wrong

HeikoKlare (approved):
Great, thanks!

PR #491, opened 2022-02-03
Enable repeated responses to the same types of user interactions in tests

### LINKED ISSUE #488: Default Response to UserInteraction during Propagation
The classes `TestUserInteraction` provides different options to define and store responses to different user interactions. The conditions of these responses are tried to be matched if an interaction occurs and if the response matches it is consumed. This allows you to check whether all predefined responses were used and therefore if all interactions occurred as expected.

However, it does not allow defining default-responses. If an interaction occurs multiple times (let's say `n` times) then one also must add the same reaction `n` times, even if the answer is every time the same. Concerning the development towards the benchmarx-framework (but also in general) it might be helpful to be able to do so. Since one does not know how many of the same interaction will occur, the current solution to realize default-responses is either:

1. Adding the same response multiple times (say 50 times) and if the default answer should be changed, clear all the remaining answers and add 50 copies of the new response. This requires an estimation of how many interactions will occur and has the disadvantage, that it will also clear all the other responses which then need to be re-added

2. During execution of a response build and add the very same response again if a "default"-flag is set which then needs to exist for every different response once.

My suggestion would be to add a second version of:
- `confirmations`
- `notificationReactions`
- `textInputs`
- `multipleChoices`
- `multipleChoiceMultipleSelections`
in `TestUserInteraction` with the prefix "default_" which contain the default answers. Then, at first the program tries to find a matching response in the default list which is not consumed, and if unsuccessful it tries to find one response in the normal list which will be consumed. A problem here is to check if multiple default responses might be triggered on the same interaction and which should be chosen in this case, but a solution might be to just take the first matching and let the user be responsible for what responses he added, much like it is currently with the normal responses.

HeikoKlare:
This PR adds functionality for applying responses to user interactions in the `TestUserInteraction` repeatedly to define default behavior for specific kinds of interactions. To this end, the builders for reponses now provide an `always` operations, such that one can write commands like:

```onTextInput[title.equals("model name")].always.respondWith("mymodel")```

Validating the occurred interactions with `assertAllInteractionsOccurred` of the `TestUserInteractions` does only consider occurences for which the exact number was specified (i.e., for which no responses that `always` apply were defined).

Fixes #488.

### DIFF
--- a/bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/TestUserInteraction.xtend
+++ b/bundles/testutils/tools.vitruv.testutils/src/tools/vitruv/testutils/TestUserInteraction.xtend
@@ -121,17 +121,28 @@ class TestUserInteraction {
 		onMultipleChoiceMultiSelection [true]
 	}
 
+	/**
+	 * Validates that all interactions, to which responses have been defined, have occurred. Otherwise, 
+	 * an {@link AssertionError} is thrown. Responses that have been defined for all occurrences of the 
+	 * same interaction by calling an {@code always} method of the builder will not be considered. 
+	 */
 	def assertAllInteractionsOccurred() {
-		val interactionsLeft = confirmations.size + notificationReactions.size + textInputs.size +
-			multipleChoices.size + multipleChoiceMultipleSelections.size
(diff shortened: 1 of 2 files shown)


tsaglam (approved):
Looks good, nothing stood out to me.

PR #493, opened 2022-02-14
Extract views into separate project

JanWittler:
Extracts views into a separate project, as suggested in #492.
1) Moves `ChangeableModelRepository`, `ChangePropagationAbortCause`, and `ChangePropagationListener` into the `tools.vitruv.framework.change` project (package `tools.vitruv.framework.change.models`).
2) Extracts the `tools.vitruv.framework.vsum.views` package and its subpackages into a separate project and adjusts the package name to `tools.vitruv.framework.views`.
3) Extracts the tests for `tools.vitruv.framework.vsum.views` into a separate project.

1 was required to prevent cyclic dependencies between the views and vsum projects. However, this also makes semantically sense as the classes are closer related to changes than to the VSUM.


JanWittler:
I am not really experienced with setting up an Eclipse project so please check the project configuration files like `Manifest.MF` in detail.
Especially for the `visibility:=reexport` modifier I was not sure whether to apply it for certain dependencies of the `tools.vitruv.framework.views` project or not. Do we have some guideline when to apply it?

tsaglam:
> Especially for the `visibility:=reexport` modifier I was not sure whether to apply it for certain dependencies of the `tools.vitruv.framework.views` project or not. Do we have some guideline when to apply it?

Re-exporting allows you to re-use the exported dependency when having a dependency on the exporting project. We do not have a guideline for that and I am not sure if there is any anywhere else. I personally see it as semantic bundling of dependencies. However, it can also obfuscate dependencies.

[OSGi in Action](https://github.com/sanigo/books/blob/master/%5BJAVA%5D%5BOSGi%20in%20Action%5D.pdf) says: "_There are few, if any, good reasons to use Require-Bundle with reexport visibility. This mechanism isn’t very modular, and using it results in brittle systems with high coupling._"

bundles/framework/tools.vitruv.framework.vsum/META-INF/MANIFEST.MF
```diff
@@ -8,12 +8,11 @@ Bundle-RequiredExecutionEnvironment: JavaSE-11
 Require-Bundle: org.apache.log4j,
  tools.vitruv.framework.domains;visibility:=reexport,
  tools.vitruv.framework.change;visibility:=reexport,
+ tools.vitruv.framework.views;visibility:=reexport,
```
tsaglam, line 11: Re-exporting here makes sense if we always need the views with `tools.vitruv.framework.vsum`. What you can do is remove the re-export statement and then check where it breaks. However, we already re-export change and domains anyways.

JanWittler: Since in the long term the VSUM should only be accessed and modified using views, I think it makes sense to re-export the views project here such that it is always visible when working with the VSUM.

HeikoKlare: Maybe that's not a good universal indicator, but I used to define re-exports for dependencies that are required to use the public API of the bundle (e.g. because a (return) parameter in a public API uses a class provided by a dependency). This also fits to the situation we have here, because, as you mentioned, using a V-SUM without views makes no sense.

tests/framework/tools.vitruv.framework.views.tests/.classpath
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11">
+		<attributes>
+			<attribute name="module" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="lib" path="lib/mockito-core-4.2.0.jar"/>
```
HeikoKlare, line 10: Having the tests using mockito moved from the `vsum` test project to the `views` test project, the library should be removed in the `vsum` project as I think it is not used there anymore. This affects the classpath file, the manifest, the pom and the library itself. Maybe there are also further dependencies in the `vsum` test project that can be removed due to the `view` test being moved to another project.

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/models/ChangeableModelRepository.java
```diff
@@ -1,4 +1,4 @@
-package tools.vitruv.framework.vsum.models;
+package tools.vitruv.framework.change.models;
```
HeikoKlare, line 1: Maybe we should have this packages called something like `propagation` instead of `models` because it is about change propagation rather than about models?

HeikoKlare (commented):
This looks like a great refactoring, thanks! I was first struggling with the change propagation interfaces being moved to the `change` project, but at second glance it makes sense to me to move those interface even from a semantic point of view and beyond the problem of cyclic dependencies. I only have a minor proposal for renaming, which we may discuss, and, if I see correctly, there needs to be some further cleanup done due to moving the `view` tests.

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/models/ChangeableModelRepository.java
HeikoKlare: As a comment in addition: The reason why the package was called `models` within the `vsum` project is that the package was supposed to contain functionality related to the models part of a V-SUM rather the views part. In the context of changes, however, this separation between models and views does not exist and, thus, does probably not make sense.

tsaglam: +1 for `propagation`

JanWittler: Seems like the naming debate was solved without me ;) Thanks for the improvement

tests/framework/tools.vitruv.framework.views.tests/.classpath
JanWittler: Thanks for pointing that out. I was able to remove all lib dependencies in that project.

HeikoKlare:
Can you execute the `views` tests within Eclipse? I have trouble executing them because the `byte-buddy` and `objenesis` libraries from the `vsum` tests project are missing.

JanWittler:
> Can you execute the `views` tests within Eclipse? I have trouble executing them because the `byte-buddy` and `objenesis` libraries from the `vsum` tests project are missing.

Thanks! I only checked for compile-time errors within Eclipse. I re-added the two libraries to the project.

HeikoKlare (approved):
Looks great, thanks!

PR #504, opened 2022-03-15
Make Commonalities generate a combined change propagation specification

HeikoKlare:
This PR makes the Commonalities language generate combined `ChangePropagationSpecifications` such that ones does not need to combine all the individual generated `ChangePropagationSpecification` on his or her own anymore.

More precisely, the language now generates one `ChangePropagationSpecification` for each direction between each pair of metamodels. These specifications are then combined in a provider class as a set of `ChangePropagationSpecifications`.


JanWittler (commented):
I really like the proposed solution and how it simplifies the code in the applications directory. However, I don't know the Commonalities language and its construction well enough to judge this PR's code quality on a semantic level.
If you think this is a critical PR which should be intensively reviewed, I will take the time to understand the Commonalities architecture. Otherwise, my review would be only syntactic feedback (which looks fine on a first rough overview).

HeikoKlare:
It's fine for me to receive syntactic feedback. Even for me, it took quite some time to understand the language architecture and it will probably not be easier when only reading the code in comparison to trying to modify it 🙂 

PS: I have converted the PR back to draft mode because there seems to be some issue with the application that I was not aware of.

JanWittler (approved):
👍🏼

PR #506, opened 2022-03-16
Fix faulty generation of deletion changes for EReference replacement

### LINKED ISSUE #505: Change monitor creates faulty deletions when replacing single-valued references
## Problem
The `ChangeMonitor` generates `DeleteEObject` changes in certain cases in which the element has actually not been deleted.

## Reproduction
If a single-valued `EReference` already references some `EObject` and this object is replaced by another `EObject` that was previously contained somewhere else, a `DeleteEObject` change for the moved `EObject` will be created, although it was not deleted.

## Reason
The reason for this issue is that the functionality in the `ChangeMonitor` to determine whether an element was deleted unintendedly ignores when an element was re-inserted into a single-valued `EReference`. This is a result of a faulty [switch statement](https://github.com/vitruv-tools/Vitruv/blob/ca41ac7346500e9dc18660a3194e6f1d7edf0138/bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/ChangeRecorder.xtend#L172-L177), in which both cases should be executed for a `ReplaceSingleValuedEReference` change but only the insertion is processed.

## Solution
Replace the switch statement by two independent, subsequent handlers.

HeikoKlare:
Fixes the faulty generation of a `DeleteEObject` change when an object is moved into a single-valued `EReference` in the `ChangeMonitor`, as documented in #505.
Provides a regression test for the scenario.

Fixes #505.


bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/ChangeRecorder.xtend
```diff
@@ -172,6 +172,8 @@ class ChangeRecorder implements AutoCloseable {
 			switch (eChange) {
 				EObjectSubtractedEChange<?> case eChange.isContainmentRemoval:
 					removedElements += eChange.oldValue
+			}
+			switch (eChange) {
 				EObjectAddedEChange<?> case eChange.isContainmentInsertion:
 					removedElements -= eChange.newValue
 			}
```
JanWittler, line 179: Maybe a matter of style but why a switch with only a single case?
Intention would be more readable for me if it was `if (eChange instanceof EObjectSubtractedEChange<?>)` and same for Added.

HeikoKlare: Done. The only drawback is that we then need two nested if statements for each case to avoid an additional type cast, but I think that's fine.

JanWittler (approved):

Issue #507, opened 2022-03-24
Missing Tests for the Insertion of Multiple Elements into an EReference


NeumannDirk:
A test class for the insertion of emelents into an `EReference` does exist [here](https://github.com/vitruv-tools/Vitruv/blob/master/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend), but it does not yet contain tests for the insertion of multiple elements at once in the form of `record [ eReference.addAll(listOfElements) ]`. This test should be included to evolve from there to future more sophisticated tests concerning planned functionality of making Change-Sequences more independent from possible modifications after recording such as deleting changes from the sequence.

JanWittler:
@NeumannDirk Can this ticket be closed due to #509 ?

NeumannDirk:
Yes

PR #509, opened 2022-03-27
Adding Tests for the Insertion of Multiple Elements into an EReference

NeumannDirk:
Adds the missing tests mentioned in https://github.com/vitruv-tools/Vitruv/issues/507#

### DIFF
--- a/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
+++ b/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
@@ -69,7 +69,7 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 	}
 	
 	@Test
-	def void testMultipleAtOnceContainment() {		
+	def void testInsertMultipleAtOnceContainment() {		
 		val Iterable<NonRoot> nonRootElements = this.createNonRootElements(5)
 		val List<EChange> result = UPR.record [
 			multiValuedContainmentEReference.addAll(nonRootElements)
@@ -84,7 +84,7 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 	}
 	
 	@Test
-	def void testMultipleAtOnceNonContainment() {
+	def void testInsertMultipleAtOnceNonContainment() {
 		val Iterable<NonRoot> nonRootElements = this.createNonRootElements(5)
 		nonContainmentHelperAddAllAsContainment(nonRootElements)
 		val List<EChange> result = UPR.record [
@@ -100,32 +100,32 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 	}
 	
 	@Test
-	def void testInsertSingleEReferenceNonContainment() {
-		testInsertInNonContainmentEReference(0)
+	def void testInsertSingleNonContainment() {
+		insertAndAssertSingleNonContainment(0)
 	}
 
 	@Test
-	def void testInsertMultipleEReferenceNonContainment() {
-		testInsertInNonContainmentEReference(0)
-		testInsertInNonContainmentEReference(1)
-		testInsertInNonContainmentEReference(2)
-		testInsertInNonContainmentEReference(1)
+	def void testInsertMultipleIterativelyNonContainment() {
+		insertAndAssertSingleNonContainment(0)
+		insertAndAssertSingleNonContainment(1)
+		insertAndAssertSingleNonContainment(2)
+		insertAndAssertSingleNonContainment(1)
 	}
 
 	@Test
-	def void testInsertSingleEReferenceContainment() {
-		testInsertInContainmentEReference(0)
+	def void testInsertSingleContainment() {
+		insertAndAssertSingleContainment(0)
 	}
 
 	@Test
-	def void testInsertMultipleEReferenceContainment() {
-		testInsertInContainmentEReference(0)
-		testInsertInContainmentEReference(1)
-		testInsertInContainmentEReference(2)
-		testInsertInContainmentEReference(1)
+	def void testInsertMultipleIterativelyContainment() {
+		insertAndAssertSingleContainment(0)
+		insertAndAssertSingleContainment(1)
+		insertAndAssertSingleContainment(2)
+		insertAndAssertSingleContainment(1)
 	}
 
-	def private testInsertInContainmentEReference(int expectedIndex) {
+	def private insertAndAssertSingleContainment(int expectedIndex) {
 		// prepare
 		uniquePersistedRoot
 		
@@ -142,7 +142,7 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 			.assertEmpty
 	}
 
-	def private testInsertInNonContainmentEReference(int expectedIndex) {
+	def private insertAndAssertSingleNonContainment(int expectedIndex) {
 		// prepare
 		val nonRoot = aet.NonRoot
 		uniquePersistedRoot => [
--- a/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
+++ b/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
@@ -18,7 +18,7 @@ import static extension tools.vitruv.framework.tests.change.util.CompoundEChange
 
 class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTransformationTest { 
 
-	def private void assertCorrectInsertionNonContainment(List<EChange> actualChanges, Pair<NonRoot, Integer>... expectedInsertions) {
+	def private void assertCorrectInsertionNonContainment(List<EChange> actualChanges, List<Pair<NonRoot, Integer>> expectedInsertions) {
 		assertEquals(actualChanges.size, expectedInsertions.size)
 		var Iterable<? extends EChange> stepwiseConsumedResult = actualChanges
 		for (insertion : expectedInsertions) {
@@ -30,7 +30,7 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 		stepwiseConsumedResult.assertChangeCount(0)
 	}
 
-	def private void assertCorrectInsertionContainment(List<EChange> actualChanges, Pair<NonRoot, Integer>... expectedInsertions) {
+	def private void assertCorrectInsertionContainment(List<EChange> actualChanges, List<Pair<NonRoot, Integer>> expectedInsertions) {
 		assertEquals(actualChanges.size, expectedInsertions.size*3)
 		var Iterable<? extends EChange> stepwiseConsumedResult = actualChanges
 		for (insertion : expectedInsertions) {
@@ -53,12 +53,11 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 	@ParameterizedTest
 	@ValueSource(ints = #[2,3,5,10])
 	def void testInsertMultipleAtOnceContainment(int count) {
-		val ArrayList<NonRoot> nonRootElements = new ArrayList<NonRoot>()
-		(0 ..< count).forEach[
-			val element = aet.NonRoot();
+		val Iterable<NonRoot> nonRootElements = (0 ..< count).map[
+			val element = aet.NonRoot()
 			element.id = it.toString()
-			nonRootElements.add(element)
-		]		
+			return element
+		]
 		val List<EChange> result = uniquePersistedRoot.record [
 			multiValuedContainmentEReference.addAll(nonRootElements)
 		]
@@ -69,9 +68,7 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 	@ParameterizedTest
 	@ValueSource(ints = #[2,3,5,10])
 	def void testInsertMultipleAtOnceNonContainment(int count) {
-		val ArrayList<NonRoot> nonRootElements = new ArrayList<NonRoot>()
-		(0 ..< count).forEach[ nonRootElements.add(aet.NonRoot()) ]
-
+		val List<NonRoot> nonRootElements = (0 ..< count).map[ aet.NonRoot() ].toList()
 		nonContainmentHelperAddAllAsContainment(nonRootElements)
 		val List<EChange> result = uniquePersistedRoot.record [
 			multiValuedNonContainmentEReference.addAll(nonRootElements)
--- a/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
+++ b/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
@@ -17,6 +17,7 @@ import static tools.vitruv.testutils.metamodels.AllElementTypesCreators.*
 
 import static extension tools.vitruv.framework.tests.change.util.AtomicEChangeAssertHelper.*
 import static extension tools.vitruv.framework.tests.change.util.CompoundEChangeAssertHelper.*
+import java.util.stream.StreamSupport
 
 class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTransformationTest { 
 
@@ -24,81 +25,61 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 	@MethodSource("provideParametersInsertMultipleAtOnceAtIndex")
 	def void testInsertMultipleAtOnceContainment(int count, int insertAt) {
 		//prepare
-		val ids = #['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j']
-		val List<NonRoot> preparationElements = ids.map[ 
-			val element = aet.NonRoot()
-			element.id = it.toString()
-			return element
-		].toList()
+		val preparationElements = (0 ..< 10).map[ aet.NonRoot() ]
 		uniquePersistedRoot.record [ multiValuedContainmentEReference.addAll(preparationElements) ]
 
 	    //test
-	    val List<NonRoot> nonRootElements = (0 ..< count).map[
+	    val nonRootElements = (0 ..< count).map[
 			val element = aet.NonRoot()
 			element.id = it.toString()
 			return element
-		].toList()
-		val List<EChange> actualChanges = uniquePersistedRoot.record [
+		].toList
+		var Iterable<? extends EChange> actualChanges = uniquePersistedRoot.record [
 			multiValuedContainmentEReference.addAll(insertAt, nonRootElements)
 		]
 
 		//assert
-		val expectedInsertions = nonRootElements.indexed().map[new Pair(value, key + insertAt)].toList
-		assertEquals(actualChanges.size, expectedInsertions.size*3)
-		var Iterable<? extends EChange> stepwiseConsumedResult = actualChanges
+		val expectedInsertions = nonRootElements.indexed().map[new Pair(value, key + insertAt)]
+		assertEquals(actualChanges.size, expectedInsertions.size*3)		
 		for (insertion : expectedInsertions) {
 			val nonRoot = insertion.key
 			val insertAtIndex = insertion.value
-			stepwiseConsumedResult = stepwiseConsumedResult
+			actualChanges = actualChanges
 				.assertCreateAndInsertNonRoot(uniquePersistedRoot, ROOT__MULTI_VALUED_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false)
 				.assertReplaceSingleValuedEAttribute(nonRoot, IDENTIFIED__ID, null, nonRoot.id, false, false)
 		}
-		stepwiseConsumedResult.assertChangeCount(0)
 	}
 
 	@ParameterizedTest
 	@MethodSource("provideParametersInsertMultipleAtOnceAtIndex")
 	def void testInsertMultipleAtOnceNonContainment(int count, int insertAt) {
 		//prepare
-		val List<NonRoot> preparationElements = (0 ..< 10).map[ aet.NonRoot() ].toList()
+		val preparationElements = (0 ..< 10).map[ aet.NonRoot() ].toList
 		uniquePersistedRoot.multiValuedContainmentEReference.addAll(preparationElements)
 		uniquePersistedRoot.record [ multiValuedNonContainmentEReference.addAll(preparationElements) ]
 
 		//test
-		val List<NonRoot> nonRootElements = (0 ..< count).map[ aet.NonRoot() ].toList()
+		val nonRootElements = (0 ..< count).map[ aet.NonRoot() ].toList
 		uniquePersistedRoot.multiValuedContainmentEReference.addAll(nonRootElements)
-		val List<EChange> actualChanges = uniquePersistedRoot.record [
+		var Iterable<? extends EChange> actualChanges = uniquePersistedRoot.record [
 			multiValuedNonContainmentEReference.addAll(insertAt, nonRootElements)
 		]
 
 		//assert
-		val expectedInsertions = nonRootElements.indexed().map[new Pair(value, key + insertAt)].toList
+		val expectedInsertions = nonRootElements.indexed().map[new Pair(value, key + insertAt)]
 		assertEquals(actualChanges.size, expectedInsertions.size)
-		var Iterable<? extends EChange> stepwiseConsumedResult = actualChanges
 		for (insertion : expectedInsertions) {
 			val nonRoot = insertion.key
 			val insertAtIndex = insertion.value
-			stepwiseConsumedResult = stepwiseConsumedResult
+			actualChanges = actualChanges
 				.assertInsertEReference(uniquePersistedRoot, ROOT__MULTI_VALUED_NON_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false, false)
 		}
-		stepwiseConsumedResult.assertChangeCount(0)
 	}
 
 	def private static Stream<Arguments> provideParametersInsertMultipleAtOnceAtIndex() {
-	    return Stream.of(
-	            Arguments.of(2, 0),
-	            Arguments.of(3, 0),
-	            Arguments.of(5, 0),
-	            Arguments.of(10, 0),
-	            Arguments.of(2, 5),
-	            Arguments.of(3, 5),
-	            Arguments.of(5, 5),
-	            Arguments.of(10, 5),
-	            Arguments.of(2, 10),
-	            Arguments.of(3, 10),
-	            Arguments.of(5, 10),
-	            Arguments.of(10, 10)
-	    );
+	    val counts = #[2,3,5,10]
+		val insertionIndexes = #[0, 5, 10]		
+		return StreamSupport.stream(insertionIndexes.map [ index | counts.map [ Arguments.of(it, index) ]].flatten.spliterator, false)		
 	}
 
 	@Test
--- a/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
+++ b/tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
@@ -1,10 +1,11 @@
 package tools.vitruv.framework.tests.change.reference
 
 import allElementTypes.NonRoot
-import allElementTypes.Root
+import java.util.ArrayList
 import java.util.List
-import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
 import tools.vitruv.framework.change.echange.EChange
 import tools.vitruv.framework.tests.change.ChangeDescription2ChangeTransformationTest
 
@@ -14,20 +15,10 @@ import static tools.vitruv.testutils.metamodels.AllElementTypesCreators.*
 
 import static extension tools.vitruv.framework.tests.change.util.AtomicEChangeAssertHelper.*
 import static extension tools.vitruv.framework.tests.change.util.CompoundEChangeAssertHelper.*
-import java.util.ArrayList
-import org.junit.jupiter.params.provider.ValueSource
-import org.junit.jupiter.params.ParameterizedTest
 
 class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTransformationTest {
 
-	private static Root UPR;
-	
-	@BeforeEach
-	def void prepareComplexTest() {
-		UPR = getUniquePersistedRoot()
-	}
-	
-	private def Iterable<NonRoot> createNonRootElements(int count) {
+	def private ArrayList<NonRoot> createNonRootElements(int count) {
 		val ArrayList<NonRoot> nonRootElements = new ArrayList<NonRoot>()
 		(0 ..< count).forEach[
 			val element = aet.NonRoot();
@@ -37,71 +28,69 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 		return nonRootElements
 	} 
 
-	def void assertFiveNonRootsNonContainment(List<EChange> actualResult, Pair<NonRoot, Integer>... expectedInsertions) {
-		assertEquals(actualResult.size, expectedInsertions.size)
-		var Iterable<? extends EChange> stepwiseConsumedResult = actualResult
+	def private void assertCorrectInsertionNonContainment(List<EChange> actualChanges, Pair<NonRoot, Integer>... expectedInsertions) {
+		assertEquals(actualChanges.size, expectedInsertions.size)
+		var Iterable<? extends EChange> stepwiseConsumedResult = actualChanges
 		for (insertion : expectedInsertions) {
 			val nonRoot = insertion.key
 			val insertAtIndex = insertion.value
 			stepwiseConsumedResult = stepwiseConsumedResult
-				.assertInsertEReference(UPR, ROOT__MULTI_VALUED_NON_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false, false)
+				.assertInsertEReference(uniquePersistedRoot, ROOT__MULTI_VALUED_NON_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false, false)
 		}
 		stepwiseConsumedResult.assertChangeCount(0)
 	}
 
-	def void assertFiveNonRootsContainment(List<EChange> actualResult, Pair<NonRoot, Integer>... expectedInsertions) {
-		assertEquals(actualResult.size, expectedInsertions.size*3)
-		var Iterable<? extends EChange> stepwiseConsumedResult = actualResult
+	def private void assertCorrectInsertionContainment(List<EChange> actualChanges, Pair<NonRoot, Integer>... expectedInsertions) {
+		assertEquals(actualChanges.size, expectedInsertions.size*3)
+		var Iterable<? extends EChange> stepwiseConsumedResult = actualChanges
 		for (insertion : expectedInsertions) {
 			val nonRoot = insertion.key
 			val insertAtIndex = insertion.value
 			stepwiseConsumedResult = stepwiseConsumedResult
-				.assertCreateAndInsertNonRoot(UPR, ROOT__MULTI_VALUED_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false)
+				.assertCreateAndInsertNonRoot(uniquePersistedRoot, ROOT__MULTI_VALUED_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false)
 				.assertReplaceSingleValuedEAttribute(nonRoot, IDENTIFIED__ID, null, nonRoot.id, false, false)
 		}
 		stepwiseConsumedResult.assertChangeCount(0)
 	}
-	
+
 	/** Elements which are to be inserted as NonContainment-EReferences have to be contained somewhere
 	 *  so they are containment-added to the unique persisted root to then be available for non containment insertion.
 	 */
-	def void nonContainmentHelperAddAllAsContainment(Iterable<NonRoot> nonRootElementsToBeContained) {
-		UPR.multiValuedContainmentEReference.addAll(nonRootElementsToBeContained)
+	def private void nonContainmentHelperAddAllAsContainment(Iterable<NonRoot> nonRootElementsToBeContained) {
+		uniquePersistedRoot.multiValuedContainmentEReference.addAll(nonRootElementsToBeContained)
 	}
-	
+
 	@ParameterizedTest
 	@ValueSource(ints = #[2,3,5,10])
-	def void testInsertMultipleAtOnceContainment(int count) {		
+	def void testInsertMultipleAtOnceContainment(int count) {
 		val Iterable<NonRoot> nonRootElements = this.createNonRootElements(5)
-		val List<EChange> result = UPR.record [
+		val List<EChange> result = uniquePersistedRoot.record [
 			multiValuedContainmentEReference.addAll(nonRootElements)
 		]
-		assertFiveNonRootsContainment(result, 
-			new Pair(nonRootElements.get(0), 0), 
-			new Pair(nonRootElements.get(1), 1), 
-			new Pair(nonRootElements.get(2), 2),
-			new Pair(nonRootElements.get(3), 3), 
-			new Pair(nonRootElements.get(4), 4)
-		)
+		val expectedInsertions = this.zipNonRootsWithExpectedInsertionIndeces(nonRootElements, #[0,1,2,3,4])
+		assertCorrectInsertionNonContainment(result, expectedInsertions)
 	}
-		
+
 	@ParameterizedTest
 	@ValueSource(ints = #[2,3,5,10])
-	def void testInsertMultipleAtOnceNonContainment(int count) {		
-		val Iterable<NonRoot> nonRootElements = this.createNonRootElements(5)		
+	def void testInsertMultipleAtOnceNonContainment(int count) {
+		val Iterable<NonRoot> nonRootElements = this.createNonRootElements(5)
 		nonContainmentHelperAddAllAsContainment(nonRootElements)
-		val List<EChange> result = UPR.record [
+		val List<EChange> result = uniquePersistedRoot.record [
 			multiValuedNonContainmentEReference.addAll(nonRootElements)
 		]
-		assertFiveNonRootsNonContainment(result, 
-			new Pair(nonRootElements.get(0), 0), 
-			new Pair(nonRootElements.get(1), 1), 
-			new Pair(nonRootElements.get(2), 2),
-			new Pair(nonRootElements.get(3), 3), 
-			new Pair(nonRootElements.get(4), 4)
-		)
+		val expectedInsertions = this.zipNonRootsWithExpectedInsertionIndeces(nonRootElements, #[0,1,2,3,4])
+		assertCorrectInsertionNonContainment(result, expectedInsertions)
 	}
-	
+
+	def private ArrayList<Pair<NonRoot, Integer>> zipNonRootsWithExpectedInsertionIndeces(Iterable<NonRoot> nonRoots, List<Integer> integers) {
+		val ArrayList<Pair<NonRoot, Integer>> expectedInsertions = new ArrayList<Pair<NonRoot, Integer>>()
+		integers.forEach[ value, index |
+			expectedInsertions.add(new Pair(nonRoots.get(index), value))
+		]
+		return expectedInsertions
+	}
+
 	@Test
 	def void testInsertSingleNonContainment() {
 		insertAndAssertSingleNonContainment(0)
@@ -128,10 +117,10 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 		insertAndAssertSingleContainment(1)
 	}
 
-	def private insertAndAssertSingleContainment(int expectedIndex) {
+	def private void insertAndAssertSingleContainment(int expectedIndex) {
 		// prepare
 		uniquePersistedRoot
-		
+
 		// test
 		val nonRoot = aet.NonRoot
 		val result = uniquePersistedRoot.record [
@@ -145,16 +134,16 @@ class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTra
 			.assertEmpty
 	}
 
-	def private insertAndAssertSingleNonContainment(int expectedIndex) {
+	def private void insertAndAssertSingleNonContainment(int expectedIndex) {
 		// prepare
 		val nonRoot = aet.NonRoot
 		uniquePersistedRoot => [
 			multiValuedContainmentEReference.add(expectedIndex, nonRoot)
 		]
 
-		// test			
+		// test
 		val result = uniquePersistedRoot.record [
-			multiValuedNonContainmentEReference.add(expectedIndex, nonRoot)			
+			multiValuedNonContainmentEReference.add(expectedIndex, nonRoot)
 		]
 
 		// assert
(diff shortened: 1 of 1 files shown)


tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,35 +1,122 @@
...
+	
+	@Test
+	def void testMultipleAtOnceContainment() {
+		val List<NonRoot> li = #[NR1, NR2, NR3, NR4, NR5]
+		val List<EChange> result = UPR.record [
+			multiValuedContainmentEReference.addAll(li)
+		]
+		assertFiveNonRootsContainment(result, new Pair(NR1, 0), new Pair(NR2, 1), new Pair(NR3, 2),
+			new Pair(NR4, 3), new Pair(NR5, 4))
+	}
```
JanWittler, line 87: While on it, we could maybe parameterize the size of the list?

Example implementation:
```Xtend

private def createNonRootElements(int count) {
  return (0 ..< count).map [
    val element = aet.NonRoot()
    element.id = it.toString() //or if there is a reason to start ids at 1 use (it + 1).toString()
    return element
  ]
}

@ParameterizedTest
@ValueSource(ints = #[2,3,5,10])
def void testInsertMultipleAtOnceContainment(int count) {
  val elements = createNonRootElements(count)
  val elementPairs = elements.indexed()
  //...
}
```

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,35 +1,122 @@
...
+	
+	@BeforeEach
+	def void prepareComplexTest() {
+		UPR = getUniquePersistedRoot()
+		NR1.id = 1.toString()
+		NR2.id = 2.toString()
+		NR3.id = 3.toString()
+		NR4.id = 4.toString()
+		NR5.id = 5.toString()
+	}
```
JanWittler, line 35: I would refactor this such that the elements are dynamically created in the complex tests. The current solution creates those elements also in test cases where they are not needed.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,35 +1,122 @@
...
+			multiValuedContainmentEReference.add(0, NR1)
+			multiValuedContainmentEReference.add(0, NR2)
+			multiValuedContainmentEReference.add(0, NR3)
+			multiValuedContainmentEReference.add(0, NR4)
+			multiValuedContainmentEReference.add(0, NR5)
+		]
+	}
+	
+	@Test
+	def void testMultipleAtOnceContainment() {
```
JanWittler, line 80: `testInsertMultipleAtOnceContainment`, naming should be consistent for all test cases.
The same for `testInsertMultipleAtOnceNonContainment`.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,35 +1,122 @@
...
+			val Pair<NonRoot, Integer> insertion = expectedInsertions.get(i)
+			val nonRoot = insertion.key
+			val insertAtIndex = insertion.value
+			stepwiseConsumedResult = stepwiseConsumedResult
+				.assertChangeCount((expectedInsertions.size-i)*3)
+				.assertCreateAndInsertNonRoot(UPR, ROOT__MULTI_VALUED_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false)
+				.assertReplaceSingleValuedEAttribute(nonRoot, IDENTIFIED__ID, null, nonRoot.id, false, false)
+		}
+		stepwiseConsumedResult.assertChangeCount(0)
+	}
```
JanWittler, line 64: These two methods are pretty similar. Maybe we can reduce the duplicated code by extracting the iteration process and providing customized validation in a closure taking the current index `i` and the `stepwiseConsumedResult`.

Anyways, test helper methods should be `private`.

JanWittler (commented):
Thanks for the improvement.
I left some minor comments regarding code quality and test coverage.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk:
```suggestion
	def (Iterable<? extends EChange>, int) => Iterable<? extends EChange> matchesIterationNonContainment(Pair<NonRoot, Integer>... expectedInsertions) {		
		[Iterable<? extends EChange> stepwiseConsumedResult, int iteration |			
			val Pair<NonRoot, Integer> insertion = expectedInsertions.get(iteration)
			val nonRoot = insertion.key
			val insertAtIndex = insertion.value			
			stepwiseConsumedResult
				.assertChangeCount((expectedInsertions.size - iteration) * 3)
				.assertCreateAndInsertNonRoot(UPR, ROOT__MULTI_VALUED_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false)
				.assertReplaceSingleValuedEAttribute(nonRoot, IDENTIFIED__ID, null, nonRoot.id, false, false)	
		]		
	}
	
	def (Iterable<? extends EChange>, int) => Iterable<? extends EChange> matchesIterationContainment(Pair<NonRoot, Integer>... expectedInsertions) {		
		[Iterable<? extends EChange> stepwiseConsumedResult, int iteration |			
			val Pair<NonRoot, Integer> insertion = expectedInsertions.get(iteration)
			val nonRoot = insertion.key
			val insertAtIndex = insertion.value			
			stepwiseConsumedResult
				.assertChangeCount(expectedInsertions.size - iteration)
				.assertInsertEReference(UPR, ROOT__MULTI_VALUED_NON_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false, false)
		]		
	}
	
	def void assertIteratively(List<EChange> actualResult, (Iterable<? extends EChange>, int) => Iterable<? extends EChange> iterationChecker, int expectedInsertionsSize) {
		var Iterable<? extends EChange> stepwiseConsumedResult = actualResult		
		for (var int i = 0; i < expectedInsertionsSize; i++) {
			stepwiseConsumedResult = iterationChecker.apply(stepwiseConsumedResult, i)
		}
		stepwiseConsumedResult.assertChangeCount(0)
	}
```

I thought about it for a while, and my suggestion above is the only good way to use closures in this case with which I came up. Please have a look at my suggestion and if this is not what you had in mind feel free to elaborate.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Refactored in [4fb1adc](https://github.com/vitruv-tools/Vitruv/pull/509/commits/4fb1adc713bbe8527391e7abd8d549df686650d9)
Bug which I only noticed later on fixed in [a283909](https://github.com/vitruv-tools/Vitruv/pull/509/commits/a283909a0acf511b40fc6c21479dae7e4bb92c9a)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Refactored in [4fb1adc](https://github.com/vitruv-tools/Vitruv/pull/509/commits/4fb1adc713bbe8527391e7abd8d549df686650d9)
Bug which I only noticed later on fixed in [a283909](https://github.com/vitruv-tools/Vitruv/pull/509/commits/a283909a0acf511b40fc6c21479dae7e4bb92c9a)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Refactored in [bc87345](https://github.com/vitruv-tools/Vitruv/pull/509/commits/bc873451618486a628c45b9feacd0563b529b7be)

NeumannDirk:
I adressed your comments and left a suggestion concerning using a closure to reuse currently duplicated code in the assertions. Feel free to double check and highlight still existing problems if needed.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,42 +1,134 @@
...
 import static extension tools.vitruv.framework.tests.change.util.AtomicEChangeAssertHelper.*
 import static extension tools.vitruv.framework.tests.change.util.CompoundEChangeAssertHelper.*
-import org.junit.jupiter.api.Test
-import static extension tools.vitruv.testutils.metamodels.AllElementTypesCreators.*
-import tools.vitruv.framework.tests.change.ChangeDescription2ChangeTransformationTest
+import java.util.ArrayList
 
 class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTransformationTest {
 
+	private static Root UPR;
```
HeikoKlare, line 21: Please use comprehensible variable names. In addition, this field should not be static as it is specific to a single test case. Finally, the field should even be removed in my opinion, as it only used to avoid parameter passing between methods.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,42 +1,134 @@
...
+	
+	@BeforeEach
+	def void prepareComplexTest() {
```
HeikoKlare, line 24: Why "complex"?

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,42 +1,134 @@
...
+		val ArrayList<NonRoot> nonRootElements = new ArrayList<NonRoot>()
+		(0 ..< count).forEach[
+			val element = aet.NonRoot();
+			element.id = it.toString()
+			nonRootElements.add(element)
+		]
+		return nonRootElements
+	} 
+
+	def void assertFiveNonRootsNonContainment(List<EChange> actualResult, Pair<NonRoot, Integer>... expectedInsertions) {
```
HeikoKlare, line 38: Why "five"? The method seems to be able to handle the given number of elements which must not necessarily be five.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,42 +1,134 @@
...
+		assertEquals(actualResult.size, expectedInsertions.size)
+		var Iterable<? extends EChange> stepwiseConsumedResult = actualResult
+		for (var int i = 0; i < expectedInsertions.size; i++) {
```
HeikoKlare, line 41: Please use meaningful names instead of "i".

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
HeikoKlare: In addition, please improve naming of method an parameters. The method does not assert non roots, but asserts specific changes. The first parameter are rather "actualChanges"than an "actualResult", as in the context of the method it is unclear to which this result relates.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,42 +1,134 @@
...
+		}
+		stepwiseConsumedResult.assertChangeCount(0)
+	}
+	
+	/** Elements which are to be inserted as NonContainment-EReferences have to be contained somewhere
+	 *  so they are containment-added to the unique persisted root to then be available for non containment insertion.
+	 */
+	def void nonContainmentHelperAddAllAsContainment(Iterable<NonRoot> nonRootElementsToBeContained) {
+		UPR.multiValuedContainmentEReference.addAll(nonRootElementsToBeContained)
+	}
```
HeikoKlare, line 72: This may not be worth a dedicated method.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,42 +1,134 @@
...
+		val List<EChange> result = UPR.record [
+			multiValuedContainmentEReference.addAll(nonRootElements)
+		]
+		assertFiveNonRootsContainment(result, 
+			new Pair(nonRootElements.get(0), 0), 
+			new Pair(nonRootElements.get(1), 1), 
+			new Pair(nonRootElements.get(2), 2),
+			new Pair(nonRootElements.get(3), 3), 
+			new Pair(nonRootElements.get(4), 4)
+		)
```
HeikoKlare, line 86: The test is hard to comprehend, as there is some initialization logic in a method with unspecified behavior (`createNonRootElements`) of which a specific behavior is implicitly assumed by the subsequent validation. Why not simply initialize the elements directly in this method and then also explicitly pass the element and the feature to the `assert...` method, so that one can read the method an completely understand what happens?

HeikoKlare (commented):
Thanks for the addition of test cases. I left some comments regarding code style.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
JanWittler: I think you missed my point here. My suggestion was to use a parameterized test such that we test lists of different lengths.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
JanWittler: +1 for removing it

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
JanWittler: Thanks for the suggestion. I had a closer look at the code and realized that what really bothered me is simply how the loop is implemented.
First, we check at the start of the method the size of the `actualResult`. Since the assert statements consume always a fixed number of elements (and also that there are enough elements left to consume), we can safely remove the `assertChangeCount` in the loop and at the end.
Then, we don't need the index anymore and can use the modern for-in loop style (`for insertion: expectedInsertions`).

```Xtend
	def void assertNonRootsNonContainment(List<EChange> actualResult, Pair<NonRoot, Integer>... expectedInsertions) {
		assertEquals(actualResult.size, expectedInsertions.size)
		var Iterable<? extends EChange> stepwiseConsumedResult = actualResult
		for (insertion: expectedInsertions) {
			val nonRoot = insertion.key
			val insertAtIndex = insertion.value
			stepwiseConsumedResult = stepwiseConsumedResult
				.assertInsertEReference(UPR, ROOT__MULTI_VALUED_NON_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false, false)
		}
	}
```

And then finally, as the method becomes pretty compact with this change, we could inline it into the actual test method as it is only used there anyways.

NeumannDirk: I added this suggestion in [58b2b46](https://github.com/vitruv-tools/Vitruv/pull/509/commits/58b2b46fb5b3a36e4457befee31354c5d4bc0d7b)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Fixed in [58b2b46](https://github.com/vitruv-tools/Vitruv/pull/509/commits/58b2b46fb5b3a36e4457befee31354c5d4bc0d7b)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Fixed in [58b2b46](https://github.com/vitruv-tools/Vitruv/pull/509/commits/58b2b46fb5b3a36e4457befee31354c5d4bc0d7b)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Fixed in [c5b749e](https://github.com/vitruv-tools/Vitruv/pull/509/commits/c5b749e484a8e654d05f78e20595baba6f1a63ef)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Fixed in [c5b749e](https://github.com/vitruv-tools/Vitruv/pull/509/commits/c5b749e484a8e654d05f78e20595baba6f1a63ef)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Tried to fix this in [f1f1e4](https://github.com/vitruv-tools/Vitruv/pull/509/commits/f1f1e4de24ea1ca11944ef0d303e8e11c7d8c026)

I am not sure if I am understanding this correctly. Your issue is that one creates this (in this case 5) NonRoot elements and that they get an ID implicitely which is later used for assertion. Is this what you mean with "specific behavior is implicitly assumed"? 

Otherwise the creation is externalized to prevent code duplication in these and future test cases. The elements are added, the framework assigns the insertion indeces and I define the expected order of the insertions and the expected assigned insertion index for each element (of which the index will be more important for the "insertAtTheEnd" and "insertAtTheFront" behaviour in the next planned addition in this area)

If I am missing what you try to point out, please let me know.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: At the moment this is only used once but it might/ will be used more often. Therefore, this way is (i think) a good version to reduce duplicated code and it is also more self-documenting so I would suggest keeping it this way. But since this is also always coupled to the creation of NonRoot elements, another option might be to pass an `boolean creationForNonContainmentTest` into the creation method which leads to adding the elements to the `multiValuedContainmentEReference` right after creation.

So what do you think:
- keeping it this way
- inline this function everywhere used
- move this functionality to the creation of the elements

NeumannDirk:
I addressed all your comments and
- Extracted the assertions of containment and non-containment tests using closures
- Made multiple code style refactorings
Please let me know what how to deal with
- open question concerning code style 
- open question concerning one possibly unclear comment about "implicitly assumed behaviour"

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
JanWittler: The intention is to get rid of the global variable (which is in general recommend). Here, this avoids accidentally preserving state across tests. Therefore, the suggestion was to remove the global variable and use local variables in the test cases where such a root element is needed. If the root element is needed in some method, it should be passed as a parameter.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
JanWittler: When removing the global `UPR` variable, this method can be removed

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Unfortunately I already worked on realizing the closure version when you commented again. I reverted my previous changes and now these are still two separate assertion methods but both are reduced by the `assertChangeCount` just as you proposed. 
This was done in [6a2a540](https://github.com/vitruv-tools/Vitruv/pull/509/commits/6a2a5402950068d9b3305870cc0bf3fe65be4def)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Obsolete by [6a2a540](https://github.com/vitruv-tools/Vitruv/pull/509/commits/6a2a5402950068d9b3305870cc0bf3fe65be4def)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
JanWittler: I think there are two points in this comment:
1) Reading the method it is not clear what is validated. More precise, I can not see that in the `assert...` method something about the element's id is checked. Thus, it would be better to adjust the method name or - as I suggested in another comment - simplify the validation and inline it.
2) It is not clear that the `createNonRootElements` method assigns ids based on the 
indexes of the elements. This could be made more precise by renaming it to something like `createNonRootElementsWithIndexBasedIdentifiers` for example. But looking at the code again we can see that the ids are actually only relevant for the non-containment case. Thus it would also be valid to inline the creation to the two test methods and do the custom id-based setup only in the non-containment test method. As the tests are all quite short, by inlining functionality a new reader can immediately understand what is happening.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
JanWittler: Please do not resolve everything immediately after adjusting something related to it.

When parameterizing the test, you should also use the `count` parameter and not keep it to the static 5.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Removed in [2659f22](https://github.com/vitruv-tools/Vitruv/pull/509/commits/2659f22c8b6b8d9615d62b0db0fea81c1721a77a)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Fixed in [2659f22](https://github.com/vitruv-tools/Vitruv/pull/509/commits/2659f22c8b6b8d9615d62b0db0fea81c1721a77a)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Fixed in [2659f22](https://github.com/vitruv-tools/Vitruv/pull/509/commits/2659f22c8b6b8d9615d62b0db0fea81c1721a77a)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,123 @@
...
+
+	@ParameterizedTest
+	@ValueSource(ints = #[2,3,5,10])
+	def void testInsertMultipleAtOnceContainment(int count) {
+		val ArrayList<NonRoot> nonRootElements = new ArrayList<NonRoot>()
+		(0 ..< count).forEach[
+			val element = aet.NonRoot();
+			element.id = it.toString()
+			nonRootElements.add(element)
+		]
```
JanWittler, line 61: Check out the `map` function for this purpose:

```suggestion
		val nonRootElements = (0 ..< count).map[
			val element = aet.NonRoot()
			element.id = it.toString()
			return element
		]		
```

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,123 @@
...
+		val expectedInsertions = this.zipNonRootsWithExpectedInsertionIndeces(nonRootElements, (0 ..< count).toList)
+		assertCorrectInsertionContainment(result, expectedInsertions)
+	}
+
+	@ParameterizedTest
+	@ValueSource(ints = #[2,3,5,10])
+	def void testInsertMultipleAtOnceNonContainment(int count) {
+		val ArrayList<NonRoot> nonRootElements = new ArrayList<NonRoot>()
+		(0 ..< count).forEach[ nonRootElements.add(aet.NonRoot()) ]
```
JanWittler, line 73: Same here
```suggestion
		val nonRootElements = (0 ..< count).map[ aet.NonRoot() ]
```

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,123 @@
...
+		assertCorrectInsertionNonContainment(result, expectedInsertions)
+	}
+
+	def private ArrayList<Pair<NonRoot, Integer>> zipNonRootsWithExpectedInsertionIndeces(Iterable<NonRoot> nonRoots, List<Integer> integers) {
+		val ArrayList<Pair<NonRoot, Integer>> expectedInsertions = new ArrayList<Pair<NonRoot, Integer>>()
+		integers.forEach[ value, index |
+			expectedInsertions.add(new Pair(nonRoots.get(index), value))
+		]
+		return expectedInsertions
+	}
```
JanWittler, line 89: As we are currently using the index of the element in the list as its expected index, this method already exists in form of the `indexed() ` method.

NeumannDirk: The `indexed()`-method returns `Pair<Integer,NonRoot>` which would work if I swap the both elements insidethe assertions. This would be ok but the bigger issue I would have with this, is that I would have to change it now and then later change it back when more functionality will be added and the expected insertion indeces do not conform anymore to the 0-based insertion order.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Changed in [f658e98](https://github.com/vitruv-tools/Vitruv/pull/509/commits/f658e98dd13c856264cf9e1e6892e7ef2407cdc2)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Changed in [f658e98](https://github.com/vitruv-tools/Vitruv/pull/509/commits/f658e98dd13c856264cf9e1e6892e7ef2407cdc2)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: "which would actually be a nice test case" -> You are right, I will add this.

There is not really another concrete example for only adding multiple elements at once. The planned scenarios are as follows.

The first is the insertion of elements at the end of a list/ EReference. This is supposed to be transformed into an insertion at index -1. So for example: 
`[a,b,c,d].insertAt(e,4)`  should return an EChange which inserts `e` at `-1`
The idea behind this is to be able to manipulate the List of Echanges after the recording. For example from the change list
`[insertAt(e,4), insertAt(f,5), insertAt(g,6)]` one cannot delete the first insertion because `[insertAt(f,5), insertAt(g,6)]` is not longer applicable for `[a,b,c,d]`, but `[insertAt(f,-1), insertAt(g,-1)]` would still be.
When these EChanges are resolved they of course become normal EMFCommands again.

The second addition is the same but for insertions of lists but at the front/ index 0. So instead of `[insertAt(a,0), insertAt(b,1), insertAt(c,2)]` the List of Ecahnges should be `[insertAt(c,0), insertAt(b,0), insertAt(a,0)]`

These two scenarios were planned in consultation with Thomas Kühn and are already prepared. We just decided to split these additions inmultiple issues/PRs. So if you prefer clean seperated additions it might be better to change it now to the best version possible which probably is using the `indexed()`-method and then later change it back.

JanWittler: The scenarios you mention don't seem to need the `zip` method neither thus I would remove it. 

[Kiss principle](https://en.wikipedia.org/wiki/KISS_principle)
Thinking about future evolutions is always a good approach but we have to be cautious not to over-engineer the system. Especially with such a large code base like Vitruv has, it can happen that files are not touched for years and at the end someone is wondering why an overly complex approach for a simple task was chosen.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
JanWittler: I think the method's body is so trivial and self-explaining - as it only adds elements to a collection - that this does not count as duplicated code. Therefore, I would vote for inlining it.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Added the mentioned test cases in [76f420b](https://github.com/vitruv-tools/Vitruv/pull/509/commits/76f420b3fc6a47c242360d802ca60731cba22f17)

Reduced the creation of comparison date for assertion to a minimum functionality in [5eeccd3](https://github.com/vitruv-tools/Vitruv/pull/509/commits/5eeccd32892e96bb9a51a81d36c1b139ab266018)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Inlinde the method in [5eeccd3](https://github.com/vitruv-tools/Vitruv/pull/509/commits/5eeccd32892e96bb9a51a81d36c1b139ab266018)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Inlined assertions as well in [31620bc](https://github.com/vitruv-tools/Vitruv/pull/509/commits/31620bc61aa7d4b52ff2f4adc5228a679da41c4e)

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
+		assertEquals(actualChanges.size, expectedInsertions.size*3)
+		var Iterable<? extends EChange> stepwiseConsumedResult = actualChanges
+		for (insertion : expectedInsertions) {
+			val nonRoot = insertion.key
+			val insertAtIndex = insertion.value
+			stepwiseConsumedResult = stepwiseConsumedResult
+				.assertCreateAndInsertNonRoot(uniquePersistedRoot, ROOT__MULTI_VALUED_CONTAINMENT_EREFERENCE, nonRoot, insertAtIndex, false)
+				.assertReplaceSingleValuedEAttribute(nonRoot, IDENTIFIED__ID, null, nonRoot.id, false, false)
+		}
+		stepwiseConsumedResult.assertChangeCount(0)
```
JanWittler, line 56: Check is not needed. Automatically valid if the overall count is valid as the later iteration always consumes a fixed number of elements.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
-import tools.vitruv.framework.tests.change.ChangeDescription2ChangeTransformationTest
 
-class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTransformationTest {
+class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTransformationTest { 
+
+	@ParameterizedTest
+	@MethodSource("provideParametersInsertMultipleAtOnceAtIndex")
+	def void testInsertMultipleAtOnceContainment(int count, int insertAt) {
+		//prepare
+		val ids = #['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j']
```
JanWittler, line 27: Why do we need ids for the filler elements?

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
+		val List<NonRoot> preparationElements = ids.map[ 
+			val element = aet.NonRoot()
+			element.id = it.toString()
+			return element
+		].toList()
```
JanWittler, line 32: Unnecessary to convert it to list

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
+		uniquePersistedRoot.record [ multiValuedContainmentEReference.addAll(preparationElements) ]
+
+	    //test
+	    val List<NonRoot> nonRootElements = (0 ..< count).map[
```
JanWittler, line 36: Unnecessary type annotation

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
 
-class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTransformationTest {
+class ChangeDescription2InsertEReferenceTest extends ChangeDescription2ChangeTransformationTest { 
+
+	@ParameterizedTest
+	@MethodSource("provideParametersInsertMultipleAtOnceAtIndex")
+	def void testInsertMultipleAtOnceContainment(int count, int insertAt) {
+		//prepare
+		val ids = #['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j']
+		val List<NonRoot> preparationElements = ids.map[
```
JanWittler, line 28: Unnecessary type annotation. `val preparationElements` is sufficient.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
+			return element
+		].toList()
+		uniquePersistedRoot.record [ multiValuedContainmentEReference.addAll(preparationElements) ]
+
+	    //test
+	    val List<NonRoot> nonRootElements = (0 ..< count).map[
+			val element = aet.NonRoot()
+			element.id = it.toString()
+			return element
+		].toList()
```
JanWittler, line 40: Unnecessary to covert it to list

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
+		val List<EChange> actualChanges = uniquePersistedRoot.record [
```
JanWittler, line 41: Unnecessary type annotation

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
+			multiValuedContainmentEReference.addAll(insertAt, nonRootElements)
+		]
+
+		//assert
+		val expectedInsertions = nonRootElements.indexed().map[new Pair(value, key + insertAt)].toList
+		assertEquals(actualChanges.size, expectedInsertions.size*3)
+		var Iterable<? extends EChange> stepwiseConsumedResult = actualChanges
```
JanWittler, line 48: Unnecessary. Iteration can be performed directly on `actualChanges`.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
```diff
@@ -1,45 +1,136 @@
...
+	            Arguments.of(2, 5),
+	            Arguments.of(3, 5),
+	            Arguments.of(5, 5),
+	            Arguments.of(10, 5),
+	            Arguments.of(2, 10),
+	            Arguments.of(3, 10),
+	            Arguments.of(5, 10),
+	            Arguments.of(10, 10)
+	    );
+	}
```
JanWittler, line 102: One possible improvement to readability would be to create the list of size of insertionIndexes explicit and compute the cross-product between them like
```Xtend
val counts = #[2,3,5,10]
val insertionIndexes = #[0, 5, 10]
return insertionIndexes [ i | counts [ Arguments.of(it, i) ]].flatten
```

JanWittler (commented):
Thank you for integrating all the suggestions. For me it looks quite good, I just left some minor (final) comments.
I only added comments to the containment test but most of them also apply to the non-containment test similarly.

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: I just thought it might be easier to debug when filler and test elements have easy to read IDs which are different from each other (letters vs numbers) but I removed the IDs

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Removed

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Removed

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Removed

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Removed

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Actually, this is necessary because `addAll(insertAt, nonRootElements)` needs a `Collection<? extends NonRoot>` and `map` only returns an `Iterable<NonRoot>`

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Removed

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Yes but in order to do so `actualChanges` has to declared as an `Iterable<? extends EChange>`

tests/framework/tools.vitruv.framework.change.tests/src/tools/vitruv/framework/tests/change/reference/ChangeDescription2InsertEReferenceTest.xtend
NeumannDirk: Thanks for the suggestion. I used this, renamed `i` to `index` and converted it to a `Stream`

HeikoKlare:
Regarding the second issue when removing the `toList` operation: This is because the "stream" operations in Xtend (`map`, `filter` etc.) are executed lazy, such that every access to the returned iterable triggers another execution of the operations. This is the reason why the elements in the test cases differ, as the `map` operation is executed multiple times and returns new elements during every execution.
To enforce immediate evaluation of the operations, you can either call `toList` afterwards or use operations that are executed immediately (see e.g. `edu.kit.ipd.sdq.commons.util.java.lang.IterableUtil.mapFixed`).

NeumannDirk:
> Regarding the second issue when removing the `toList` operation: This is because the "stream" operations in Xtend (`map`, `filter` etc.) are executed lazy, such that every access to the returned iterable triggers another execution of the operations. This is the reason why the elements in the test cases differ, as the `map` operation is executed multiple times and returns new elements during every execution. To enforce immediate evaluation of the operations, you can either call `toList` afterwards or use operations that are executed immediately (see e.g. `edu.kit.ipd.sdq.commons.util.java.lang.IterableUtil.mapFixed`).

Thank you for the clearification. I now use `mapfix` instead of `map` and `toList`. I think the PR should then be complete unless you have anything else?

JanWittler (approved):
Looks fine to me 👍🏼

JanWittler:
I rebased the branch to integrate the latest changes such that this PR can be merged

PR #511, opened 2022-04-01
Adapt to use local maven build option in GitHub actions

### LINKED ISSUE #510: Add local build Maven profile
In the GitHub actions, currently, to build against the local build of the core and domains repositories `sed` is used to replace the update site location with a reference to a local build.
This is hard to understand and error-prone.
Furthermore, when testing local changes against the domains or applications repository, the same workaround has to be used to reference the local copy instead of the update site.

Both issues can be solved by adding a _local build_ Maven profile. The suggested change is to add a profile with a configurable path to the local version instead of to the update site. This change should not modify the default behavior but only occur when explicitly enabled.

JanWittler:
Adapts the GitHub actions to use the local Maven builds.
Closes #510

As part of this PR, the option to execute builds with local references should be documented in the Wiki.

⚠️ Requires [Vitruv-Domains#115](https://github.com/vitruv-tools/Vitruv-Domains-ComponentBasedSystems/pull/115) and [Vitruv-Applications#188](https://github.com/vitruv-tools/Vitruv-Applications-ComponentBasedSystems/pull/188) to be merged first ⚠️

### DIFF
--- a/.github/workflows/applicationsvalidation.yml
+++ b/.github/workflows/applicationsvalidation.yml
@@ -60,7 +60,7 @@ jobs:
         working-directory: ./cbsdomains
         run: >
             ./mvnw -B -U clean verify
-            -Dvitruv.framework.path=${maven.multiModuleProjectDirectory}/../framework
+            -Dvitruv.framework.path="\${maven.multiModuleProjectDirectory}/../framework"
             -Dstyle.color=always
             -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
             -Dorg.slf4j.simpleLogger.log.org.eclipse.tycho.core.resolver.DefaultTychoResolver=warn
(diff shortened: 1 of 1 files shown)


HeikoKlare (approved):
👍 Thanks!

HeikoKlare:
We have to use the `xvfb-action` for both the framework and the applications because these projects contain test cases in which a non-headless Eclipse is started for which an X server is required. The domains do not require that and thus run without it.

It seems that the `xvfb-run` script processes the arguments in a different way than running the `mvnw` script plainly, so we probably need to keep these different formats for defining the paths.
At least when using the property in a local setup (i.e., on your local machine, not in GitHub Actions) this will not be an issue, as you can simply run maven also for the framework and the applications because there will be some display server already running.

JanWittler:
Thanks for clarifying that.

JanWittler:
Added Wiki documentation: https://github.com/vitruv-tools/Vitruv/wiki/Build-and-Continuous-Integration

PR #513, opened 2022-04-14
Bidirectional execution in test domains

HeikoKlare:
Transformations should/will usually be executed bidirectionally. The framework tests (Reactions, Commonalities, V-SUM) tests currently only execute the test change propagation specifications in one direction. With this PR, their execution is performed in both directions until a fixed point is reached.

This PR makes the following changes:
* Enables transitive change propagation in test domains, such that Reactins and Commonalities tests are executed with transitive change propagation
* Updates the Reactions in the Reactions tests to allow their execution in a bidirectional setting. 
* Updates the Commonalities tests to conform to the bidirectional execution of generated transformations.
* Updates test change propagation specification in V-SUM tests to properly handle bidirectional execution.

### DIFF
--- a/tests/dsls/tools.vitruv.dsls.reactions.tests/src/tools/vitruv/dsls/reactions/tests/ReactionsExecutionTest.xtend
+++ b/tests/dsls/tools.vitruv.dsls.reactions.tests/src/tools/vitruv/dsls/reactions/tests/ReactionsExecutionTest.xtend
@@ -11,6 +11,8 @@ import tools.vitruv.testutils.TestProject
 import org.junit.jupiter.api.TestInstance
 import edu.kit.ipd.sdq.activextendannotations.Lazy
 import static com.google.common.base.Preconditions.checkNotNull
+import org.junit.jupiter.api.BeforeEach
+import tools.vitruv.testutils.domains.AllElementTypesDomainProvider
 
 @ExtendWith(InjectionExtension)
 @InjectWith(ReactionsLanguageInjectorProvider)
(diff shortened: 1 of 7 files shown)


tests/dsls/tools.vitruv.dsls.reactions.tests/src/tools/vitruv/dsls/reactions/tests/ReactionsExecutionTest.xtend
```diff
@@ -26,6 +28,11 @@ abstract class ReactionsExecutionTest extends VitruvApplicationTest {
 		]
 	)
 
+	@BeforeEach
+	def void patchDomain() {
+		new AllElementTypesDomainProvider().domain.enableTransitiveChangePropagation()
+	}
+
```
JanWittler, line 35: If the follow-up PR removes the `AllElementTypesDomainProvider` anyways and enables transitive propagation by default, why not set `shouldTransitivelyPropagateChanges` to `true` by default already in this PR? Then we could get rid of this piece of code and thus avoid having to clean it up in the following PR.

HeikoKlare: My intent was to have a "clean" state after this PR, so that the test domains do not behave differently (one that propagates changes transitively and the others that do not). But since there is the common super class `VitruvTestDomain`, we can of course enable transitive propagation for all test domains in that class by default.

JanWittler (approved):

PR #515, opened 2022-04-14
Improve default state-based change resolution strategy

JanWittler:
This PR adds the following improvements:

- Adds customization option to the default state-based change resolution strategy to enable / disable matching based on identifiers (defaults to _when available_)
- Fixed identifier-based comparison for resources with `xmi:id` tags (links to [Vitruv-Applications#184](https://github.com/vitruv-tools/Vitruv-Applications-ComponentBasedSystems/issues/184)). Ids are only copied within the scope of the view and not when creating a view from the V-SUM. This is done as the ids there don't have any semantic as they are not detected by our change recorder.
- Adds a factory method for modifying views using a state-based view

### Details on `xmi:id`

As explained in [Vitruv-Applications#184](https://github.com/vitruv-tools/Vitruv-Applications-ComponentBasedSystems/issues/184), the ids may not be stored in the `EObject`s but in the `Resource`. These ids are not copied when using the default `EcoreUtil` copy functionality. However, these ids are used by `EMFCompare` for identifier-based matching. This resulted in incorrect matching as the created resource copies did not contain the same ids as the original resources. This occurred in the change-deriving view when creating the reference state resource set as well as in the default state-based change derivation strategy when creating a resource copy to apply the derived differences to (to convert them to `EChange`s).

### DIFF
--- a/tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/changederivation/BasicStateChangePropagationTest.xtend
+++ b/tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/changederivation/BasicStateChangePropagationTest.xtend
@@ -27,10 +27,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
 		getModelURI("Test.allElementTypes")
 	}
 
-	@ParameterizedTest(name = "{1}")
+	@ParameterizedTest()
 	@DisplayName("create new resource and calculate state-based difference")
 	@MethodSource("strategiesToTest")
-	def void createNewResource(StateBasedChangeResolutionStrategy strategyToTest, String name) {
+	def void createNewResource(StateBasedChangeResolutionStrategy strategyToTest) {
 		val modelResource = new ResourceSetImpl().createResource(testUri) => [
 			contents += aet.Root => [
 				id = "Root"
@@ -52,10 +52,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
 		assertThat(validationResourceSet.resources.get(0), containsModelOf(modelResource))
 	}
 
-	@ParameterizedTest(name = "{1}")
+	@ParameterizedTest()
 	@DisplayName("delete existing resource and calculate state-based difference")
 	@MethodSource("strategiesToTest")
-	def void deleteResource(StateBasedChangeResolutionStrategy strategyToTest, String name) {
+	def void deleteResource(StateBasedChangeResolutionStrategy strategyToTest) {
 		val modelResource = new Capture<Resource>
 		resourceSet.record [
 			createResource(testUri) => [
@@ -79,10 +79,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
 		assertTrue(validationResourceSet.resources.get(0).contents.empty)
 	}
 
-	@ParameterizedTest(name = "{1}")
+	@ParameterizedTest()
 	@DisplayName("replace root element and calculate state-based difference")
 	@MethodSource("strategiesToTest")
-	def void replaceRootElement(StateBasedChangeResolutionStrategy strategyToTest, String name) {
+	def void replaceRootElement(StateBasedChangeResolutionStrategy strategyToTest) {
 		val modelResource = new Capture<Resource>
 		resourceSet.record [
 			createResource(testUri) => [
@@ -110,10 +110,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
 		assertThat(validationResourceSet.resources.get(0), containsModelOf(-modelResource))
 	}
 
-	@ParameterizedTest(name = "{1}")
+	@ParameterizedTest()
 	@DisplayName("change a root element property and calculate state-based difference")
 	@MethodSource("strategiesToTest")
-	def void changeRootElementFeature(StateBasedChangeResolutionStrategy strategyToTest, String name) {
+	def void changeRootElementFeature(StateBasedChangeResolutionStrategy strategyToTest) {
 		val modelResource = new Capture<Resource>
 		val root = aet.Root
 		resourceSet.record [
@@ -141,10 +141,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
 		assertThat(validationResourceSet.resources.get(0), containsModelOf(-modelResource))
 	}
 	
-	@ParameterizedTest(name = "{1}")
+	@ParameterizedTest()
     @DisplayName("change a root element's id and calculate state-based difference")
     @MethodSource("strategiesToTest")
-    def void changeRootElementId(DefaultStateBasedChangeResolutionStrategy strategyToTest, String name) {
+    def void changeRootElementId(DefaultStateBasedChangeResolutionStrategy strategyToTest) {
         val modelResource = new Capture<Resource>
         val root = aet.Root
         resourceSet.record [
@@ -188,10 +188,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
         assertThat(validationResourceSet.resources.get(0), containsModelOf(-modelResource))
     }
 
-	@ParameterizedTest(name = "{1}")
+	@ParameterizedTest()
 	@DisplayName("change a non-root element property and calculate state-based difference")
 	@MethodSource("strategiesToTest")
-	def void changeNonRootElementFeature(StateBasedChangeResolutionStrategy strategyToTest, String name) {
+	def void changeNonRootElementFeature(StateBasedChangeResolutionStrategy strategyToTest) {
 		val modelResource = new Capture<Resource>
 		val root = aet.Root
 		val containedRoot = aet.Root
@@ -224,10 +224,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
 		assertThat(validationResourceSet.resources.get(0), containsModelOf(-modelResource))
 	}
 	
-	@ParameterizedTest(name = "{1}")
+	@ParameterizedTest()
     @DisplayName("change a non-root element's id and calculate state-based difference")
     @MethodSource("strategiesToTest")
-    def void changeNonRootElementId(DefaultStateBasedChangeResolutionStrategy strategyToTest, String name) {
+    def void changeNonRootElementId(DefaultStateBasedChangeResolutionStrategy strategyToTest) {
         val modelResource = new Capture<Resource>
         val root = aet.Root
         val containedRoot = aet.Root
@@ -276,10 +276,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
         assertThat(validationResourceSet.resources.get(0), containsModelOf(-modelResource))
     }
 
-	@ParameterizedTest(name = "{1}")
+	@ParameterizedTest()
 	@DisplayName("move a resource to new location and calculate state-based difference")
 	@MethodSource("strategiesToTest")
-	def void moveResource(StateBasedChangeResolutionStrategy strategyToTest, String name) {
+	def void moveResource(StateBasedChangeResolutionStrategy strategyToTest) {
 		val modelResource = new Capture<Resource>
 		val root = aet.Root
 		resourceSet.record [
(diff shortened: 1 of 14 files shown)


JanWittler:
I am struggling with how to implement proper tests for the `DefaultStateBasedChangeResolutionStrategy`. [Currently](https://github.com/vitruv-tools/Vitruv/blob/master/tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/changederivation/BasicStateChangePropagationTest.xtend), we are requiring the derived changes to be identical to the recorded ones. However, this cannot be guaranteed and only works because the tests are quite simple and until now always used identifier-based matching.
For similarity-based matching, we cannot make any assumptions about how the change sequence looks but only that our view should be in the expected state after applying the derived changes.
For identity-based matching, we can assume that matches are always computed correctly and thus keep making strong assumptions about the change sequence. There is one edge case, namely changing an element's id. In this case we can only require that there is a delete and a create change (as an element with a changed id must not match with the old element in identifier-based matching).

Thus I see two options for similarity-based matching: Relax the tests to only validate the resulting model or keep the requirements on the change sequence (which currently works for almost all test cases). The former is a weaker validation with our atomic changes, the latter option could lead to problems in the future when the `DefaultStateBasedChangeResolutionStrategy` is modified to e.g. use an alternate model comparison framework.
For identify-based matching I would preserve the current logic with requiring the exact change sequence.

What are your opinions on this topic @HeikoKlare @tsaglam ?

HeikoKlare:
First of all: I think it is reasonable to compare model states instead of changes.
Some thoughts on that:

I would phrase the question a little different, because it should actually not be about what we _can_ test but about what we _should_ test, so: What is the contract we expect the strategy to fulfill and how should we test that?
We may have not made that contract explicit, but, in general, this contract should require that the result when applying the changes delivered by the strategy is the same as the final state given to the strategy to derive the changes. This would be an argument for only comparing the model states.

I could, however, imagine that there are some scenarios in which we expect a specific change sequence to be derived and then actually have to validate the changes. For example, we might expect the strategy to sequentialize Java changes in a proper way, i.e., first generating classifier changes, then genering signature changes and finally genering method implementations, as other orders may be invalid. That will, however, a specific test for Java views and not for the general strategy.

We could, of course, also think about keeping the change comparison until it fails (e.g. because we alter the model comparison framework). That may, however, be confusing, as one has to know that it may be valid that the tests will fail when changing the tested functionality, as they are actually too strict w.r.t. the contract expected from the functionality. In short: we should only test behavior what we actually expect and if we agree that we not require specific changes but only a specific result from applying them, we may only test the latter.

tsaglam:
> We could, of course, also think about keeping the change comparison until it fails (e.g. because we alter the model comparison framework). That may, however, be confusing, as one has to know that it may be valid that the tests will fail when changing the tested functionality, as they are actually too strict w.r.t. the contract expected from the functionality. In short: we should only test behavior what we actually expect and if we agree that we not require specific changes but only a specific result from applying them, we may only test the latter.

This was one point I was thinking about as well, regarding maintainability it is not ideal to check the changes when the expected behavior is related to the resulting states. If we want to keep it as is and avoid confusion in the future, we could also add comments to the test cases as a "dirty" hotfix that mention that framework changes can change the changes thus leading to test failures.

JanWittler:
Thanks for your comments.
Based on your feedback I would propose the following solution:
For the scope of this PR keep the logic of validating the obtained change sequence as-is and open an issue to refactor the tests such that state validation is applied. I would prefer to not include this refactoring already in this PR as it requires quite some effort, plus this PR does not introduce any stronger constraints than there were before. The refactoring would require state comparison for the UML and PCM mockup models. Additionally, we should rethink the state validation of the basic tests as the state comparison is using EMF Compare and using the same tool for validation that is validated is not the best ([who watches the watchers](https://en.wikipedia.org/wiki/Quis_custodiet_ipsos_custodes%3F)).

I had to disable two test cases of the UML mock which did not yield the expected change sequences. Except of that, all tests succeed with all 3 id matching strategies. For the newly added identifier changing test cases, I already provided a more generic change sequence validation that should be valid for any underlying strategy (at least for the id-based matching).

tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/changederivation/BasicStateChangePropagationTest.xtend
```diff
@@ -134,10 +140,55 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
...
+        switch (strategyToTest.useIdentifiers) {
+            case ONLY,
+            case WHEN_AVAILABLE: {
+                assertEquals(1, changes.EChanges.filter(CreateEObject).map[affectedEObject].filter(Root).filter [
+                    id == "Root2"
+                ].size)
+                assertEquals(1, changes.EChanges.filter(DeleteEObject).map[affectedEObject].filter(Root).filter [
+                    id == "Root"
+                ].size)
+            }
```
JanWittler, line 175: Eclipse is always giving me the warning about `map[affectedEObject]` that
> CreateEObject is a raw type. References to generic type CreateEObject<A>
   should be parameterized

but I could not find out how to do this in a filter pipeline. Can someone help out? What I want to assert here is that a new element of type `Root` with id "Root2" was created and an existing element of type `Root` with id "Root" was deleted.

This can always be asserted for any strategy matching by identifiers as the old id is not present anymore resp. the new id was not present before and thus no matches should be created for such elements.

HeikoKlare: Xtend-Streams with generics is a weird thing. I have also struggled with that problem several times, but I do not have a good solution in mind.
Two not very good options are:
1. Convert the filtered iterable into a correctly typed list:
```Xtend
val Iterable<CreateEObject<?>> createEObjectChanges = newArrayList(change.EChanges.filter(CreateEObject))
val affectedObjects = createEObjectChanges.map[affectedEObject]
```
2. Implement `filter` by hand:
```Xtend
change.EChanges.fold(new ArrayList, [list, element | if (element instanceof CreateEObject<?>) list += element; list]).map[affectedEObject]
```
Both alternatives will have some performance overhead.

JanWittler: Thanks, that is a little bit disappointing 😄 
I've opted in for the first solution in https://github.com/vitruv-tools/Vitruv/pull/515/commits/761d1c2ce5e8f823a7b36bf7f49952930c35466c

tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/changederivation/BasicStateChangePropagationTest.xtend
```diff
@@ -25,9 +27,10 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
 		getModelURI("Test.allElementTypes")
 	}
 
-	@Test
+	@ParameterizedTest(name = "{1}")
 	@DisplayName("create new resource and calculate state-based difference")
-	def void createNewResource() {
+	@MethodSource("strategiesToTest")
+	def void createNewResource(StateBasedChangeResolutionStrategy strategyToTest, String name) {
```
JanWittler, line 33: I don't really like this approach with passing a second argument just to have the tests named. However, I did not find a solution to have a computed name based on the first value.
A different approach would be to just pass the UseIdentifier enum values. This would result in automatically readable names but I do not like this approach as it forces me to instantiate the strategy in the test each time and does not allow to add new strategies at a single location in code.

HeikoKlare: It may be possible to define a custom `DisplayNameGenerator`, which transforms the given strategy argument into an according string. But I have not used that yet, so I am not sure about it.

JanWittler: Thanks for the pointer. Turns out there is an even simpler solution by using the `Named` interface.
https://github.com/vitruv-tools/Vitruv/pull/515/commits/93b3f765cee5e37348c51cdc038091399c7c2070

HeikoKlare (commented):
I only have some minor comments regarding code style. In addition, the `BasicStateChangePropagationTest` is formatted inconsistently (whitespace vs. tab indentations), which should probably be unified before merging this PR.

tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/util/XmiIdEdgeCaseTest.java
```diff
@@ -0,0 +1,119 @@
...
+	private Path testProjectFolder;
+	private XMLResource umlModel;
+	private Map<String, EObject> expectedIdMapping;
+
+	@BeforeEach
+	void setup(@TestProject Path testProjectFolder) {
+		this.testProjectFolder = testProjectFolder;
+		resourceSet = withGlobalFactories(new ResourceSetImpl());
+
+		this.umlModel = (XMLResource) resourceSet.createResource(getModelURI("my.uml_mockup"));
```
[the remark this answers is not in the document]
JanWittler: Property was removed completely due to refactoring the resource creation into a specific method in https://github.com/vitruv-tools/Vitruv/pull/515/commits/270725419c68bb4f522b868b99a3bceac2fe036d

tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/util/XmiIdEdgeCaseTest.java
```diff
@@ -0,0 +1,119 @@
...
+		uPackage2.getClasses().add(uClass2);
+
+		expectedIdMapping = Map.of( //
+				"package-1", uPackage1, //
+				"package-2", uPackage2, //
+				"class-1", uClass1, //
+				"class-2", uClass2 //
+		);
+		expectedIdMapping.forEach((id, obj) -> umlModel.setID(obj, id));
+	}
```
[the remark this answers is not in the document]
JanWittler: Done in https://github.com/vitruv-tools/Vitruv/pull/515/commits/270725419c68bb4f522b868b99a3bceac2fe036d

tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/util/XmiIdEdgeCaseTest.java
```diff
@@ -0,0 +1,119 @@
...
+		assertNotNull(copiedModel2, "copy for uml model 2 is missing");
+		validateIds(copiedModel, expectedIdMapping);
+		validateIds(copiedModel2, expectedIdMapping2);
+	}
+
+	private void validateIds(Resource copiedResource, Map<String, EObject> expectedIdMapping) {
+		expectedIdMapping.forEach((id, object) -> {
+			EObject copiedObject = copiedResource.getEObject(id);
+			assertNotNull(copiedObject, "could not find element with id " + id);
+			assertEquals(((Identified) object).getId(), ((Identified) copiedObject).getId());
```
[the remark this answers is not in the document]
JanWittler: Done in https://github.com/vitruv-tools/Vitruv/pull/515/commits/270725419c68bb4f522b868b99a3bceac2fe036d

tests/framework/tools.vitruv.framework.views.tests/src/tools/vitruv/framework/views/changederivation/BasicStateChangePropagationTest.xtend
```diff
@@ -134,10 +140,58 @@ class BasicStateChangePropagationTest extends StateChangePropagationTest {
...
+
+        val validationResourceSet = new ResourceSetImpl().withGlobalFactories()
+        val oldState = validationResourceSet.getResource(testUri, true)
+        val changes = strategyToTest.getChangeSequenceBetween(-modelResource, oldState)
+        switch (strategyToTest.useIdentifiers) {
+            case ONLY,
+            case WHEN_AVAILABLE: {
+                val Iterable<DeleteEObject<?>> deleteChanges = newArrayList(changes.EChanges.filter(DeleteEObject))
+                assertEquals(1, deleteChanges.size)
+                assertTrue(deleteChanges.head.affectedEObject instanceof Root)
```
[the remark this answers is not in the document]
JanWittler: Done in https://github.com/vitruv-tools/Vitruv/pull/515/commits/270725419c68bb4f522b868b99a3bceac2fe036d

HeikoKlare (approved):
Thanks for the improvements!

tsaglam (approved):

Issue #512, opened 2022-04-06
Adding Functionality to Insert Elements in an EReference at the Last Position


NeumannDirk:
Currently, the recording of the insertion of an element into an `EReference` yields an `InsertEReference` which contains a `newValue` and an `index`. One of the most common cases is the insertion of an element at the end of the `EReference` in the case of which the `index` is equal to the number of elements in the `EReference` during the insertion. The resulting `InsertEReference` is then dependent on the correct number of elements in the `EReference` to be applied. This is espacially hindering in the case of the insertion of multiple elements at once. were the `InsertEReference`s from the recording have to be applied in the correct order and none of the `InsertEReference`s can be left out (except the last one).

An addition to this functionality should be implemented to allow for the manipulation of the order in which the `InsertEReference`s 
 are applied and the deletion of certain `InsertEReference`s without loosing the applicability of the remaining `InsertEReference`s.

To give an example:
Currently, inserting a list of elements `[d,e,f]` into an `EReference` which already contains the elements `[a,b,c]` returns a Sequence of `InsertEReference`s `[insert(d, 4), insert(e, 5), insert(f, 6)]`
- The three insertions can be applied in this order only
- Deleting an insertion makes the successing insertions unapplicable

A better version might be if the recording creates the following list: `[insert(d, -1), insert(e, -1), insert(f, -1)]` where `-1` means "at the end of the `EReference`" or "at `ereference.size()`"
- The three insertions can still be applied in this order but can be shuffled as well
- Deleting an insertion does not effect the remaining insertions

NeumannDirk:
I am planning to address this issue in a PullRequest in a timely manner.

JanWittler:
I am not really sure about this change. Should this apply only for changes within the VSUM or also for changes propagated from some view?
If it is a change in the VSUM, it solves -- as you said -- the problem that I can leave out changes of the sequence when constructing a view and still correctly insert the element. However, this only solves the problem for this one particular case: insertion at the end. An insertion at any other position could still not be handled if a change inserting an element at a lower index before is left out. Furthermore, would a change with a fixed index to append an element still be valid or do we require every such change to have the `-1` index? The primer would introduce ambiguity to the change recording, the latter would make strong requirements about internals.
[...]

PR #517, opened 2022-04-26
Adding Functionality to Insert Elements in an EReference at the Last Position

NeumannDirk:
Addressing #512 this PR adds the mentioned functionality as well as the corresponding tests.


bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/NotificationToEChangeConverter.xtend
```diff
@@ -195,15 +199,33 @@ package final class NotificationToEChangeConverter {
...
+				createInsertReferenceChange(notifierModelElement, reference, value, INSERT_LAST_INDEX).
+					surroundWithCreateAndFeatureChangesIfNecessary()
+			]
+		} else {
+			listOfNewValues.flatMapFixedIndexed [ index, value |
+				createInsertReferenceChange(notifierModelElement, reference, value, initialIndex + index).
+					surroundWithCreateAndFeatureChangesIfNecessary()
+			]
+		}
 	}
```
HeikoKlare, line 229: Please refactor this, such that the common logic for calculating the index is encapsulated in a dedicated method taking, e.g., the initial index of added elements, the number elements in list after addition and the number of added elements, and returning either the given index or -1. Then the existing code can be kept as is except for calculating the (initial) index via a method call.

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/command/ApplyBackwardCommandSwitch.xtend
```diff
@@ -32,6 +32,8 @@ import static extension tools.vitruv.framework.change.echange.command.ChangeComm
 package class ApplyBackwardCommandSwitch {
 	static val Logger logger = Logger.getLogger(ApplyBackwardCommandSwitch)
 	
+	final static int INSERT_LAST_INDEX = -1;
```
HeikoKlare, line 35: This index is specified multiple times, but not where it belongs to. Why not provide it somewhere at the `EChange` implementation, since it specifies the semantics of `EChanges` with index `-1`.

HeikoKlare, line 35: This index is specified multiple times, but not where it belongs to. Why not provide it somewhere at the `EChange` implementation, since it specifies the semantics of `EChanges` with index `-1`.

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/command/ApplyBackwardCommandSwitch.xtend
```diff
@@ -91,7 +93,14 @@ package class ApplyBackwardCommandSwitch {
...
 		} else {
-			return #[new RemoveAtCommand(editingDomain, change.affectedEObject, change.affectedFeature, change.newValue, change.index)]
+			/* Do not manipulate the change itself, but to convert them to the predefined EMF-Commands,
+			 * change the -1 index back to the last index in the list.
+			 */
+			var index = change.index
+			if(change.index == INSERT_LAST_INDEX) {
+				index = (change.affectedEObject.eGet(change.affectedFeature) as List<EObject>).size() - 1
+			}
+			return #[new RemoveAtCommand(editingDomain, change.affectedEObject, change.affectedFeature, change.newValue, index)]
```
HeikoKlare, line 103: Why is this functionality implemented in the command generation rather than the command application (i.e., within `RemoveAtCommand`)? In consequence, `-1` is only properly handled but here, e.g., not in the opposite `ApplyForwardCommandSwitch`.

HeikoKlare, line 103: Why is this functionality implemented in the command generation rather than the command application (i.e., within `RemoveAtCommand`)? In consequence, `-1` is only properly handled but here, e.g., not in the opposite `ApplyForwardCommandSwitch`.

HeikoKlare (changes requested):
Thanks for providing functionality for handling changes with index `-1`. See detailed comments for required improvements.

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/command/ApplyBackwardCommandSwitch.xtend
NeumannDirk: Fixed in [0cf41bb](https://github.com/vitruv-tools/Vitruv/pull/517/commits/0cf41bb2c3c93922073c9c3f269d70ddc58a23c9)

NeumannDirk: Fixed in [0cf41bb](https://github.com/vitruv-tools/Vitruv/pull/517/commits/0cf41bb2c3c93922073c9c3f269d70ddc58a23c9)

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/command/ApplyBackwardCommandSwitch.xtend
NeumannDirk: Fixed in [e9e2963](https://github.com/vitruv-tools/Vitruv/pull/517/commits/e9e2963ff37b66af06c71551fea7a7767971f8e4)

NeumannDirk: Fixed in [e9e2963](https://github.com/vitruv-tools/Vitruv/pull/517/commits/e9e2963ff37b66af06c71551fea7a7767971f8e4)

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/NotificationToEChangeConverter.xtend
NeumannDirk: Possible solution implemented in [37f79a2](https://github.com/vitruv-tools/Vitruv/pull/517/commits/37f79a25820060ec2a76e6fbc964a6127699278d). Please let me know if this is what you had in mind.

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/command/RemoveAtCommand.xtend
```diff
@@ -51,16 +52,23 @@ class RemoveAtCommand extends RemoveCommand {
...
+			
 		if (!result) {
 			return false;
 		}
-		// Check if get(index) == object		
-		return ownerList.get(index).equals(collection.get(0))
+		// Check if get(index) == object
+		val elementAtIndex = if (index == EChangeUtil.LAST_POSITION_INDEX) ownerList.get(ownerList.size - 1) else ownerList.get(index)
+		return elementAtIndex.equals(collection.get(0))
 	}
```
HeikoKlare, line 72: This functionality is quite compilcated. Why not have a method for the index property, which either gives the actual index if it is >= 0 or calculates the actual index from the `ownerList` size if it is -1. Then the functionality can be kept as is except for accessing an index property method instead of the index itself.

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/NotificationToEChangeConverter.xtend
```diff
@@ -195,16 +198,35 @@ package final class NotificationToEChangeConverter {
...
 				surroundWithCreateAndFeatureChangesIfNecessary()
 		]
 	}
+	
+	private def int getEChangeIndex(int originalFirstInsertionPosition, int counterOfCurrentInsertion, int sizeAfterInsertion, int numOfNewValues) {
+		return if (originalFirstInsertionPosition == sizeAfterInsertion - numOfNewValues) 
+			EChangeUtil.LAST_POSITION_INDEX 
+		else 
+			originalFirstInsertionPosition + counterOfCurrentInsertion
+	}
```
HeikoKlare, line 229: This modification is not what I meant in the previous comment. I meant the factor out the _common functionality_ (of `handleInsertReference` and `handleMultiInsertReference`), as they both need to handle the case that the (first) index of the inserted element(s) may be -1. That calculation should be done by a utility method. And it should be donein the same way, as currently the `handleInsertReference` method checks for the element to be contained at a specific position (which is actually invalid, as the element may have already been in the list at the last index but was inserted at another index) whereas the `handleMultiInsertReference` method properly uses only the index.

HeikoKlare (changes requested):
See my comments

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/NotificationToEChangeConverter.xtend
HeikoKlare: See my comment in the new review.

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/command/RemoveAtCommand.xtend
JanWittler: Here we should also think about whether we want to store the `resolvedIndex` - i.e. once compute it in `prepare` and then only use that index in `doExecute` instead of the case distinction - or always compute it on the fly.
From my understanding the index must not change after `prepare` got called so it probably is even safer to store it, but best would be to check the `Command` documentation in detail for that.

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/NotificationToEChangeConverter.xtend
```diff
@@ -195,16 +198,35 @@ package final class NotificationToEChangeConverter {
 	}
 
 	private def Iterable<? extends EChange> handleInsertReference(extension NotificationInfo notification) {
-		createInsertReferenceChange(notifierModelElement, reference, newModelElementValue, position).
+		/* If the new value was just inserted and especially at the last position,
+		 * then the index can be changed to -1.
+		 */
+		var newIndex = position
+		var eRef = notifierModelElement.eGet(reference) as List<EObject>
```
JanWittler, line 205: is this type safe?

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/EChangeUtil.xtend
```diff
@@ -13,6 +13,9 @@ import edu.kit.ipd.sdq.activextendannotations.Utility
  */
 @Utility
 class EChangeUtil {
+	
+	public final static int LAST_POSITION_INDEX = -1;
```
JanWittler, line 17: documentation

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/command/RemoveAtCommand.xtend
NeumannDirk: I agree with Heiko to extract the functionality to get the resolved index when you need it and reuse this function.
I am not so sure about Jans suggestion to store this resolved index in advance since the resolved index should be dependent on the current state of the `ownerList`. If the `RemoveAtCommand` is created and the resolved index is stored during the creation and then the `ownerList` is manipulated somewhere else then the resolvedIndex does not represent the original intention anymore of this `RemoveAtCommmand`. Am I right?
Concerning `prepare`, the documentation says:
"It is very convenient to use prepare, as it is guaranteed to be called only once just before canExecute is to be tested. It can be implemented to create any additional commands that need to be executed, and the result it yields becomes the permanent cached return value for canExecute."
So the assumption here might be that there wont be any modifications on the `ownerList` after `prepare` was called, but as I said I am not so sure about the timespan between Command creation and `prepare` call.


Therefore I would suggest the change as in [cdbe99f](https://github.com/vitruv-tools/Vitruv/pull/517/commits/cdbe99f3a38ced0e591ac18019c8172579fecbc8)

JanWittler: Thanks for looking up the documentation. I also checked and in `AbstractCommand` it says
> Called at most once in {@link #canExecute} to give the command an opportunity to ready itself for execution. The returned value is stored in {@link #canExecute}. In other words, you can override this method to initialize and to yield a cached value for the all subsequent calls to canExecute.

As the computation of the index is not very computation intensive I think it is fine to not cache it but rather use the computation method every time as you suggested.

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/NotificationToEChangeConverter.xtend
HeikoKlare: Many methods in this class make assumptions on the `NotificationInfo` that is passed to them, i.e., that they are only called when the `NotificationInfo` encapsulates an event they are supposed to handle, such as `InsertReference` assuming to be called for inserting an element into a multi-valued reference. None of the methods either documents or validates that assumption. Since these methods encapsulate internal, private functionality, I think it is valid to make these assumptions instead of checking them in every method producing overhead. But maybe we need to have some documentation for that. However, we could have an issue for that and add it to the whole class instead of adding it for a single method in this PR. What do you think?

JanWittler: Ah, thanks for pointing that out. I was just unsure whether the typecast to List is covered by the assumption made for that method.
I also think that it is fine to make that assumptions. We could add documentation but I do not really see the priority for that, so maybe just postpone it for another PR.

NeumannDirk: As Heiko pointed out, there are assumptions made on the `NotificationInfo` and the return value of the extension method`getNotifierModelElement` which just returns an `EObject`. The only thing one can know for sure is that the `getReference` does not return null but an actual `EReference`.
In [cca9543](https://github.com/vitruv-tools/Vitruv/pull/517/commits/cca9543748c5ef855517080b5d3f7c665000764e), I changed the cast to `as Iterable<Object>` which is a somewhat weaker assumption.

bundles/framework/tools.vitruv.framework.change/src/tools/vitruv/framework/change/recording/NotificationToEChangeConverter.xtend
NeumannDirk: I corrected the check as you mentioned and extracted it as method in [cca9543](https://github.com/vitruv-tools/Vitruv/pull/517/commits/cca9543748c5ef855517080b5d3f7c665000764e)

bundles/framework/tools.vitruv.framework.change.echange/src/tools/vitruv/framework/change/echange/EChangeUtil.xtend
NeumannDirk: I added a short documentation in [85cd7e3](https://github.com/vitruv-tools/Vitruv/pull/517/commits/85cd7e305174b62c410c56539dc776f5d1b1d44c). I pointed out that this is only used for the addition/ insertion and for reference by naming them explicitely. If this gets adopted for `EAttributes` as well in the future this should be edited.

JanWittler:
In general, it would be beneficial for an `open` PR to be 
1) up-to-date with the `master` branch and
2) have all tests running
You can test your local changes against the applications as documented in the [Wiki](https://github.com/vitruv-tools/Vitruv/wiki/Build-and-Continuous-Integration#github-actions).

If the PR is not yet ready but you still want to communicate your changes, you can make it a `draft` PR.

HeikoKlare:
I propose to close this PR, as it requires much more sophisticated functionality to achieve its goal.

NeumannDirk:
Thank you for having a look into this, since I had some issues with building and verifying with maven locally. It seems that you found the problem with this PR and therefore the reason why the concept of the "-1"-insertion won't work for now. Therefore, we should close this PR. At least this attempt is documented in this history, and it might be helpful for future attempts to implement a functionality for insertions at the last positions / adding elements without stating a certain index.

PR #518, opened 2022-05-02
Minor framework cleanup

JanWittler:
1) Removes some outdated documentation regarding UUIDs.
2) Removes exception hiding in `ApplyEChangeSwitch` as all test cases work without catching the referenced exception.


HeikoKlare (approved):
Thanks for cleaning up 👍

PR #520, opened 2022-05-12
Refactored VaVe metamodel

SofiaAnanieva:
This is a draft pull request for the refactored VaVe metamodel. It is the first step of VaVe refactorings (branch variability). Further changes may follow.


JanWittler:
I had a first look at the model, didn't find anything problematic or surprising to me. Some minor comments:
- relations should follow camelCaseNotationStyle (e.g. `rootFeature` instead of `rootfeature`)
- there are multiple to-many relationships with a singular name, should be plural
- why are `Feature.treeconstraint` and `TreeConstraint.feature` not setup as opposites?

Besides that we obviously need documentation but as the PR is still a draft, I guess this will come in the progress. We should also adjust the main Readme to document the new `variability` top level element in the Project Structure section.

SofiaAnanieva:
Thanks a lot for your helpful comments! I just added a commit to this draft pull request to address them. 
Regarding your third point about why `Feature.treeconstraint` and `TreeConstraint.feature` are not opposites: That was because the former represented the child tree constraints of a feature and the letter represented the child features of a tree constraint, and neither represents the parents (i.e., neither of the two were bi-directional relations). I addressed this now by renaming the references to `TreeConstraint.childFeatures` and `Feature.childTreeConstraints` and giving them parent-opposites to make them bi-directional.

JanWittler:
So if I want to iterate over some feature tree I'd have to go `childFeatures` -> `childTreeConstraints` -> `childFeatures` and so on (i.e. alternate between `Feature` and `TreeConstraint`) ? This seems to me as if we should extract this into a proper tree data structure.

JanWittler:
Closed due to inactivity

PR #522, opened 2022-06-08
Remove unnecessary dependency in views test project

JanWittler:
This PR removes one unnecessary dependency of the `tools.vitruv.framework.views.tests` target that makes the build in Eclipse fail for Eclipse 03/22 on macOS.


tsaglam:
Wait, this dependency is the source of the failure? 👀

tsaglam (approved):

HeikoKlare (approved):

PR #525, opened 2022-06-10
GitHub actions update

JanWittler:
Updates the GitHub actions primarily to avoid the undocumented `maven.multiModuleProjectDirectory` variable.

> This variable is only used in the mvn script and is neither documented nor intended for public used (internal implementation detail). So I strongly recommend not to use it. Otherwise things might break in the future.

[Quote source](https://stackoverflow.com/questions/29778262/what-is-maven-multimoduleprojectdirectory-used-for)

Additionally, versions of used GitHub actions are updated.
As of `setup-java@v2` the Java distribution needs to be specified. I chose to preserve the default value of v1 which was `Zulu` but I am open for discussion to use other Java distributions.

### DIFF
--- a/.github/workflows/applicationsvalidation.yml
+++ b/.github/workflows/applicationsvalidation.yml
@@ -61,7 +61,7 @@ jobs:
         working-directory: ./cbsdomains
         run: >
             ./mvnw -B -U clean verify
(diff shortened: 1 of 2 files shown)


tsaglam (approved):
Looks good! This is by the way one of the things the dependabot would do for us, when we enable it.

PR #531, opened 2022-06-18
Reference latest kit-sdq updatesites

HeikoKlare:
Replaces the references to the composite kit-sdq updatesites with references of the latest versions. This avoid accidental resolution of artifacts from old versions and improves build times as not all versions of the updatesites have to be resolved.

### DIFF
--- a/releng/tools.vitruv.parent/pom.xml
+++ b/releng/tools.vitruv.parent/pom.xml
@@ -16,17 +16,17 @@
 		<repository>
 			<id>Demo Metamodels</id>
 			<layout>p2</layout>
-			<url>https://kit-sdq.github.io/updatesite/release/metamodels/demo/</url>
+			<url>https://kit-sdq.github.io/updatesite/release/metamodels/demo/latest/</url>
 		</repository>
 		<repository>
 			<id>SDQ Commons</id>
 			<layout>p2</layout>
-			<url>https://kit-sdq.github.io/updatesite/release/commons/</url>
+			<url>https://kit-sdq.github.io/updatesite/release/commons/latest/</url>
 		</repository>
 		<repository>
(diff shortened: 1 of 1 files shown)


tsaglam (approved):

JanWittler:
Just to understand the change: By referencing the latest updatesite, we only have access to the latest release? I.e. if we decide in the future to constrain the kit-sdq to some specific version which might not be the latest for whatever reason, we'd have to rollback these changes to access such version?

HeikoKlare:
Yes, that's true. Actually, we have three possibilities to reference the updatesite for a repository:
1. Use the composite updatesite, which provides all versions (like before this PR). If we then want to reference an older version of a dependency, we have to define that version in the bundle dependencies (which we currently don't do for any dependency). The drawback of this approach is that build times increase as the updatesite for every version provided by the composite updatesite has to be retrieved and inspected.
2. Use a specific version of the updatesite (like before #530). Then we have to adapt the updatesite in the build specification whenever we require a new version. The drawback of this approach is that we need to recognize that a new version is available on our own and then adapt the build specification.
3. Use the lastest version of the updatesite (like proposed in this PR). Then we have the guarantee to be always up-to-date, which fits to the behavior of our other dependencies (in particular from the Eclipse updatesite). The drawback is that we cannot simply set a specific dependency version in the build specification anymore (despite replacing the `lastest` with the proper version again) but have to put the version in the bundle dependencies.

Actually, another option would be to have a composite updatesite for each major release. Then we could use the latest version of a specific major release, as every new minor release should not break anything.

JanWittler (approved):
Thanks for the clarification. I think for now it is a consistent solution with our other dependencies. We just have to keep in mind to adjust it again if we do proper dependency version management in the future.

PR #540, opened 2022-06-23
Remove unused functionality from domains

HeikoKlare:
This is the first PR to finally get rid of domains. It removes much of the functionality of domains, which are not required anymore.
In particular it removes the following:
* Definition of transitive propagation, which has been moved to the `ChangePropagator`
* Default load and save options
* Visibility of domains for declaring demo domains
* Functionality to determine whether an element belongs to a domain
* Definition of `furtherRootPackages` without a proper definition what such a "further" package is

To still provide functionality for transitive change propagation, the `VirtualModel` now provides the ability to enable or disable it for a complete V-SUM (by passing it to the underlying `ChangePropagator`).


bundles/tools.vitruv.framework.vsum/src/tools/vitruv/framework/vsum/VirtualModel.xtend
```diff
@@ -7,4 +7,10 @@ import tools.vitruv.framework.views.ViewTypeProvider
 
 interface VirtualModel extends ChangeableModelRepository, ViewProvider, ViewTypeProvider {
 	def Path getFolder()
+	
+	/**
+	 * Enables or disables transitive propagation of changes when passed to {@link #propagateChange(VitruviusChange)}.
+	 * Transitive propagation is enabled by default.
+	 */
+	def void setTransitivePropagationEnabled(boolean enabled)
```
JanWittler, line 15: `setTransitiveChangePropagationEnabled` to be more precise?

HeikoKlare: Yes, that makes sense.

JanWittler (approved):

PR #541, opened 2022-06-24
Remove domains

HeikoKlare:
This PR completely removes the obsolete `VitruvDomains`.
* Removes domains from `VirtualModel`
* Makes project builder and their applicators independent from domains (as they are actually an instrument for views, not for domains, still has to be renamed and refactored as documented in #542)
* Replaces the selection of domains in wizard with selection of builder applicators
* Removes the `tools.vitruv.framework.domains` and `tools.vitruv.domains.demo` projects and according features or feature entries


bundles/tools.vitruv.domains.emf.ui/src/tools/vitruv/domains/emf/builder/VitruvEmfBuilder.java
```diff
@@ -49,6 +49,10 @@ public class VitruvEmfBuilder extends VitruvProjectBuilder {
     public VitruvEmfBuilder() {
         this(null, null, null);
     }
+    
+    private boolean monitorsModelWithFileExtension(String fileExtension) {
+    	return fileExtension != null && fileExtension != "" && (getMonitoredFileExtensions().isEmpty() || getMonitoredFileExtensions().contains(fileExtension));
```
JanWittler, line 54: If `getMonitoredFileExtensions()` returns an empty set, the method will return true for any file extension provided. Seems not to be intended by the name of the method?

HeikoKlare: I have to admit that the name is a bit misleading. Unfortunately, I have not documented that there is some other place in the code saying that if no file extensions are specified, models with all file extensions will be monitored (so the empty set is a kind of magic value like -1). Maybe this behavior should be stated at the `getMonitoredFileExtensions()` method?

JanWittler: Ah ok. Yes, that would probably be helpful. Is this "magic value" only for the EmfBuilder or for any builder? For the latter it would make sense to move the `monitorsModelWithFileExtension` method to the parent class where the `getMonitoredFileExtensions` is implemented.

HeikoKlare: Since the only other builder is for Java and thus only monitors Java files, it will probably not be used anywhere else. In fact, we have to at least partly reimplement the builder anyway to make it properly usable for views.

HeikoKlare: I have taken another look at it and moved the method to the general `VitruvProjectBuilder`, as you have proposed. The semantics is also valid for the `VitruvJavaBuilder`. I have also added documentation for the behavior.

bundles/tools.vitruv.framework.domains.ui/src/tools/vitruv/framework/domains/ui/builder/VitruvProjectBuilderApplicator.xtend
```diff
@@ -2,11 +2,15 @@ package tools.vitruv.framework.domains.ui.builder
 
 import org.eclipse.core.resources.IProject
 import java.util.Set
-import tools.vitruv.framework.domains.VitruvDomain
 import com.google.common.collect.Sets
 import java.nio.file.Path
 
 interface VitruvProjectBuilderApplicator {
+	/**
+	 * Returns the name of builder
```
JanWittler, line 10:
```suggestion
	 * Returns the name of the builder
```

bundles/tools.vitruv.framework.domains.ui/src/tools/vitruv/framework/domains/ui/builder/VitruvProjectBuilderApplicatorImpl.xtend
```diff
@@ -37,7 +41,7 @@ final class VitruvProjectBuilderApplicatorImpl implements VitruvProjectBuilderAp
 	override void addBuilder(IProject project, Path virtualModelFolder, Set<String> fileExtensions) {
 		checkArgument(project !== null, "Project must not be null")
 		checkArgument(virtualModelFolder !== null, "Virtual model folder must not be null")
-		checkArgument(!fileExtensions.nullOrEmpty, "File extensions of builder must not be null or empty")
+		checkArgument(fileExtensions !== null, "File extensions of builder must not be")
```
JanWittler, line 44:
```suggestion
		checkArgument(fileExtensions !== null, "File extensions of builder must not be null")
```

bundles/tools.vitruv.framework.domains.ui/src/tools/vitruv/framework/domains/ui/builder/VitruvProjectBuilderRegistry.xtend
```diff
@@ -1,88 +1,61 @@
...
 	public static val INSTANCE = new VitruvProjectBuilderRegistry
-	static val runtimeRegisteredBuilderIds = new HashMap<VitruvDomain, Set<String>>
+	static val runtimeRegisteredBuilderIds = new HashSet<Pair<String, String>>
 	
 	new() {
 	}
 
 	/**
-	 * Registers project builder with the given ID for the given {@link VitruvDomain}.
+	 * Registers project builder with the given name and ID.
```
JanWittler, line 21:
```suggestion
	 * Registers a project builder with the given name and ID.
```

JanWittler (approved):
Looks fine. There were just minor mistakes in documentation

PR #558, opened 2022-08-17
Eclipse 2022-06

JanWittler:
Improve dependencies for Eclipse 2022-06 updating parent POM to 2.0.0. New parent POM defaults to non-platform tests, thus marker files can be removed.

Due to breaking changes, we bump the development version from 2.1.0 to 3.0.0

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <extensions>
   <extension>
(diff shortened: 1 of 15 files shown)


HeikoKlare (approved):
May we have the branches for the Eclipse 2022-06 update named equally, such that the validation produces usable results? I don't know if we can easily rename a branch for which already a PR exists, otherwise I may also rename the branches in the other repositories and set up new PRs.

PR #581, opened 2023-03-05
Update parent pom to 2.1.2 and Maven to 3.9.0

JanWittler:
Draft PR as parent pom 2.1.2 is not released yet. Dependency updates should be tested in Eclipse 2023-03.


larsk21:
superseded by #587

PR #594, opened 2023-07-04
Make `EChange` generic over its content type

JanWittler:
Adapts to https://github.com/vitruv-tools/Vitruv-Change/issues/59


JanWittler:
@vitruv-tools/maintainers Due to the new type safety for changes, it becomes apparent that the tool adapters are not working anymore. Besides probably many flaws there, right now they don't compile anymore because there is no functionality to transform from the recorded `VitruviusChange<EObject>` to the required `VitruviusChange<Uuid>`. This problem is not introduced by the PR but only uncovered.

I suggest to make the Tool Adapters legacy. I don't think it is worth it to hack some UUID resolver into the adapters to make it work. The proper way of fixing them would be migrate to the View-API but that would require rewriting most of the code anyways.

PR #640, opened 2024-12-10
Add module descriptions to readme

larsk21:


larsk21:
What exactly do we need the `applications` module for?

Issue #687, opened 2025-02-11
Change the visibility of this constructor to "protected"


arlange:
https://sonarcloud.io/project/issues?open=AZNj0v8eZ5M1X0Sd407-&id=vitruv-tools_Vitruv

arlange:
Pull request is still open. Please fix.

PR #690, opened 2025-02-12
Updated Access Modifiers

Muhammad-Tahir-KIT:


TomWerm (approved):

larsk21:
Please link the related issues in the bar on the right ("Development") or by using one of the GitHub keywords, such as `closes`. If the issues should not be closed by the PR, you can still mention them like this: "see #687".

arlange (changes requested):
there are 2 constructors in the class. Both need the protected instead of public access modifier.

arlange (changes requested):
protected access modifier for all constructors in this class

arlange (approved):

PR #741, opened 2025-06-04
bump multiple dependencies from dependabot

arlange:


TomWerm:
I think we could also bump up the maven build parent in this PR. What do you think?

TomWerm (approved):

PR #771, opened 2025-07-18
ResourceRepositoryImpl-issues

TsotneMikadze:
fixed:
https://sonarcloud.io/project/issues?open=AZSRLUCaLqnSWQ2gsXVu&id=vitruv-tools_Vitruv

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,20 +54,19 @@ jobs:
         java-version: '17'
         check-latest: true
         cache: 'maven'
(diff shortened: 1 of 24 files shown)


TsotneMikadze:
i changed it

ma-mirzaei (changes requested):
Please move the messages file to the up level and create a package named Message. In this package we need to find two classes. The address should be like below.
.../vsum/internal/messages/ErrorMessages.java

PR #861, opened 2026-06-16
Java 21 ready

arlange:


arlange:
after the change from Java 17 to 21 in the workflows, all builds should work properly.

PR #893, opened 2026-09-13
Implement changes of PR in vitruv-change

jeanpmathes:
This implements the methods added to interfaces in PR https://github.com/vitruv-tools/Vitruv-Change/pull/560 . Additionally, the VirtualModelBuilder has received a few new methods that allow setting the level of change propagation specifications (see PR in vitruv change) and creating view types that require the ViewTypeProvider.


jeanpmathes:
The build will only be possible when the changes have arrived in vitruv-change. Maybe there is a way to do the build differently?

arlange:
> The build will only be possible when the changes have arrived in vitruv-change. Maybe there is a way to do the build differently?

No, we just have to wait for the nightly build to be finished, once the PR on the Vitruv-Change side has finished.

